PR #1, opened 2018-06-18
DBZ-20 Moving debezium-connector-oracle from main to incubator repo

gunnarmorling:
Hey @jpechane, here's the PR for adding the Oracle connector to this new incubator repo.

Can you do a santity check (in conjunction with https://github.com/debezium/debezium/pull/535) and merge if you think it's alright? IMO it is, of course there are many things left to be done, but I think we can take those via incremental PRs from here.


jpechane:
@gunnarmorling Good work, thanks!

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectionFactory.java
```diff
@@ -0,0 +1,29 @@
...
+    @Override
+    public Connection connect(JdbcConfiguration config) throws SQLException {
+        String hostName = config.getHostname();
+        int port = config.getPort();
+        String database = config.getDatabase();
+        String user = config.getUser();
+        String password = config.getPassword();
+
+        return DriverManager.getConnection(
+              "jdbc:oracle:oci:@" + hostName + ":" + port + "/" + database, user, password
```
jchipmunk: Why do you use `oci` instead of `thin` driver implementation?

gunnarmorling: Good question; IIRC, XStreams doesn't work with thin.

PR #2, opened 2018-06-21
DBZ-749 Modifications to support Maven release plug-in

jpechane:


pom.xml
```diff
@@ -37,6 +39,37 @@
                 <artifactId>debezium-connector-oracle</artifactId>
                 <version>${project.version}</version>
             </dependency>
+            <dependency>
```
gunnarmorling: Wouldn't it suffice to specify `${version.debezium}` in the reference to the _debezium-parent_ POM?

gunnarmorling:
Yes, no problem. Can you only file an issue so we don't forget about it.
Thanks!

PR #3, opened 2018-06-28
DBZ-707 Removing redundant dependency info

gunnarmorling:


jpechane:
@gunnarmorling Command `mvn release:clean release:prepare -DreleaseVersion=0.8.0.Final -Dtag=v0.8.0.Final -DdevelopmentVersion=0.9.0-SNAPSHOT -DpushChanges=false '-Darguments=-DskipTests -DskipITs -Passembly -Dversion.debezium=0.8.0.Final -Poracle' -Dversion.debezium=0.8.0.Final -Poracle` switches to non-interactive mode that prevents automated release

gunnarmorling:
@jpechane, ok, I understand now what's going on. The artifact versions are set to `$project.version` in the parent's dependency management block. When the build of _debezium-incubator_ is executed, this property then is resolved to the version of that project (incubator) and not the parent.

All in all I think it's best if we leave things as is then. It's a bit redundant, but not worth any more efforts IMO. I'll close this PR and the issue. Thanks for investigating!

PR #6, opened 2018-07-05
DBZ-725 Removing DB name from topic names

gunnarmorling:
https://issues.jboss.org/browse/DBZ-725


jpechane:

PR #7, opened 2018-07-09
DBZ-791 Using "c##xstrm" user in tests (non-admin user);

gunnarmorling:
Also using non-Flashback related SELECT for getting current SCN.

https://issues.jboss.org/browse/DBZ-791

@jpechane 2nd part of that change :)


gunnarmorling:
@jpechane Pushed one more commit to close results more safely. This should go into the Final, too. Best is to view the last commit ignoring whitespaces, then it's apparent that nothing much has changed at all.

jpechane:

PR #8, opened 2018-07-10
DBZ-720 Extracting HistorizedRelationalSnapshotChangeEventSource

gunnarmorling:
Requires https://github.com/debezium/debezium/pull/569

@jpechane Here's the counterpart to the core PR.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
@@ -6,132 +6,72 @@
 package io.debezium.connector.oracle;
 
 import java.sql.Clob;
-import java.sql.Connection;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
-import java.util.HashSet;
 import java.util.Set;
 import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import io.debezium.pipeline.source.spi.SnapshotChangeEventSource;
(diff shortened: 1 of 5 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
```diff
@@ -6,229 +6,144 @@
...
-        }
+    @Override
+    protected Set<TableId> getAllTableIds(SnapshotContext ctx) throws Exception {
+        return jdbcConnection.readTableNames(ctx.catalogName, null, null, new String[] {"TABLE"} );
     }
 
-    private void determineCapturedTables(SnapshotContext ctx) throws SQLException {
-        Set<TableId> allTableIds = jdbcConnection.readTableNames(ctx.catalogName, null, null, new String[] {"TABLE"} );
+    @Override
+    protected boolean lockTables(ChangeEventSourceContext sourceContext, SnapshotContext snapshotContext) throws SQLException {
```
jpechane: Would it be possible to pull this method up and introduce an abstract method lockTable()?

jpechane:
@gunnarmorling One comment add, otherwise it is good to go

gunnarmorling:
I was wondering that, too. But then I thought some DBs might be able to
lock all tables at once (e.g. MySQL), so it seemed preferable to let
implementations iterate themselves. But we might re-consider based on your
other question around a "lockDatabase()" method. I think we should adjust
this step by step, as we add more implementations. Knowing the specific
needs, it'll be easier than trying to design it upfront.

jpechane:
@gunnarmorling ok, I'll probably do both thinks when I use it for SQL server :-)

jpechane:
@gunnarmorling Appled, thanks!

PR #11, opened 2018-07-12
DBZ-720 snapshotting

gunnarmorling:
https://issues.jboss.org/browse/DBZ-720

@jpechane And here's the snapshotting implementation for Oracle. Based on top of #8, only the last commit is new.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
@@ -13,6 +13,7 @@ import org.apache.kafka.connect.errors.ConnectException;
 
 import io.debezium.config.CommonConnectorConfig;
 import io.debezium.config.Configuration;
+import io.debezium.config.EnumeratedValue;
 import io.debezium.config.Field;
 import io.debezium.document.Document;
 import io.debezium.jdbc.JdbcConfiguration;
@@ -81,6 +82,16 @@ public class OracleConnectorConfig extends RelationalDatabaseConnectorConfig {
             .withValidation(Field::isRequired)
             .withDescription("Name of the XStream Out server to connect to.");
 
+    public static final Field SNAPSHOT_MODE = Field.create("snapshot.mode")
+            .withDisplayName("Snapshot mode")
+            .withEnum(SnapshotMode.class, SnapshotMode.INITIAL)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.LOW)
+            .withDescription("The criteria for running a snapshot upon startup of the connector. "
+                    + "Options include: "
+                    + "'initial' (the default) to specify the connector should run a snapshot only when no offsets are available for the logical server name; "
+                    + "'initial_schema_only' to specify the connector should run a snapshot of the schema when no offsets are available for the logical server name. ");
+
     /**
      * The set of {@link Field}s defined as part of this configuration.
      */
@@ -89,6 +100,7 @@ public class OracleConnectorConfig extends RelationalDatabaseConnectorConfig {
             DATABASE_NAME,
             PDB_NAME,
             XSTREAM_SERVER_NAME,
+            SNAPSHOT_MODE,
             RelationalDatabaseConnectorConfig.TABLE_WHITELIST,
             RelationalDatabaseConnectorConfig.TABLE_BLACKLIST,
             RelationalDatabaseConnectorConfig.TABLE_IGNORE_BUILTIN,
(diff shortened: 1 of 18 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -81,6 +82,16 @@
             .withValidation(Field::isRequired)
             .withDescription("Name of the XStream Out server to connect to.");
 
+    public static final Field SNAPSHOT_MODE = Field.create("snapshot.mode")
```
jpechane: This should belong to the parent class

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/SnapshotChangeRecordEmitter.java
```diff
@@ -0,0 +1,43 @@
...
+import io.debezium.relational.RelationalChangeRecordEmitter;
+import io.debezium.util.Clock;
+import oracle.streams.RowLCR;
+
+/**
+ * Emits change data based on a single {@link RowLCR} event.
+ *
+ * @author Gunnar Morling
+ */
+public class SnapshotChangeRecordEmitter extends RelationalChangeRecordEmitter {
```
jpechane: Is this class specific to oracle?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
gunnarmorling: I was wondering that, too, but then I figured that the actual snapshotting modes are different between the different connectors. Hence the idea of `SnapshottingTask`, based on its specific snapshotting modes and the previous offset (if any), a specific sub-class returns this descriptor which indicates whether a snapshot of schema, data or both is needed.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/SnapshotChangeRecordEmitter.java
gunnarmorling: I suppose that it's not. Let's move it to core once it's proven with SQL Server.

gunnarmorling:
@jpechane Rebased to current master.

jpechane:
@gunnarmorling Could yoy please check `OracleDatabaseSchema`? It seems that an unused imports crept in in a previous commits which fails build.
Also `StreamingDatatypesIT` and `SnapshotDatatypesIT` does not work as records created as part of `io.debezium.connector.oracle.OracleConnectorIT.shouldTakeSnapshot()` seems to stay in database and are delivered to the aforementioned tests.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleOffsetContext.java
```diff
@@ -64,4 +123,62 @@ public void setTransactionId(String transactionId) {
...
+        }
+
+        @Override
+        public Map<String, ?> getPartition() {
+            return Collections.singletonMap(SERVER_PARTITION_KEY, logicalName);
+        }
+
+        @Override
+        public OffsetContext load(Map<String, ?> offset) {
+            long scn = (long) offset.get(SourceInfo.SCN_KEY);
```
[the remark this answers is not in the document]
gunnarmorling: Interesting, I'm not geting this failure. But I'll change it regardless.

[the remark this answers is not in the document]
jpechane: The one I get is
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project debezium-connector-oracle: Compilation failure
[ERROR] /home/jpechane/sources/debezium/debezium-incubator/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleOffsetContext.java:[177,41] incompatible types: capture#1 of ? cannot be converted to long
```

gunnarmorling:
Out of curiousity, which JDK version are you on?

jpechane:
```
openjdk version "1.8.0_171"
OpenJDK Runtime Environment (build 1.8.0_171-8u171-b11-0ubuntu0.16.04.1-b11)
OpenJDK 64-Bit Server VM (build 25.171-b11, mixed mode)
```

gunnarmorling:
@jpechane Ok, I've pushed one more commit which I hope fixes all the issues. Build times (well, testing times) are killing me, though :(

jpechane:
@gunnarmorling Thanks a lot it works now, applied!

PR #14, opened 2018-07-17
DBZ-720 Adjusting snapshotting test

gunnarmorling:
https://issues.jboss.org/browse/DBZ-720

@jpechane This adjusts the `SnapshotDatatypesIT` so to actually test the (data) snapshotting. I think with that we can consider DBZ-720 as complete (unless you still see anything missing).


jpechane:

PR #17, opened 2018-07-17
DBZ-801 Parameterizing ChangeEventQueue<DataChangeEvent>

gunnarmorling:
Requires https://github.com/debezium/debezium/pull/579

https://issues.jboss.org/browse/DBZ-801


jpechane:

PR #20, opened 2018-07-23
DBZ-804 Emit NUMBER columns as Int32/Int64 if precision and scale allow

gunnarmorling:
https://issues.jboss.org/browse/DBZ-804

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
@@ -25,10 +25,16 @@ import io.debezium.relational.TableId;
 import io.debezium.relational.Tables;
 import io.debezium.relational.Tables.ColumnNameFilter;
 import io.debezium.relational.Tables.TableFilter;
+import oracle.jdbc.OracleTypes;
 
 public class OracleConnection extends JdbcConnection {
 
     private final static Logger LOGGER = LoggerFactory.getLogger(OracleConnection.class);
+
+    /**
+     * Returned by column metadata in Oracle if no scale is set;
+     */
+    private static final int ORACLE_UNSET_SCALE = -127;
 
     public OracleConnection(Configuration config, ConnectionFactory connectionFactory) {
         super(config, connectionFactory);
@@ -116,6 +122,19 @@ public class OracleConnection extends JdbcConnection {
                                     .scale(null)
                                     .create()
                                 );
+                    }
+                    // NUMBER columns without scale value have it set to -127 instead of null;
+                    // let's rectify that
+                    else if (column.jdbcType() == OracleTypes.NUMBER) {
+                        column.scale()
+                            .filter(s -> s == ORACLE_UNSET_SCALE)
+                            .ifPresent(s -> {
+                                editor.addColumn(
+                                        column.edit()
+                                            .scale(null)
+                                            .create()
+                                        );
+                            });
                     }
                 }
                 tables.overwriteTable(editor.create());
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDdlParser.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDdlParser.java
@@ -324,6 +324,9 @@ public class OracleDdlParser implements DdlParser {
             if (precisionPart.numeric().size() > 1) {
                 columnEditor.scale(Integer.valueOf(precisionPart.numeric(1).getText()));
             }
+            else if (precisionPart.numeric_negative() != null) {
+                columnEditor.scale(Integer.valueOf(precisionPart.numeric_negative().getText()));
+            }
             else {
                 columnEditor.scale(0);
             }
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDdlParser.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDdlParser.java
@@ -179,7 +179,9 @@ public class OracleDdlParser implements DdlParser {
                     }
                     columnEditor.length(6);
                 }
-                else if (ctx.datatype().native_datatype_element().VARCHAR2() != null) {
+                // VARCHAR is the same as VARCHAR2 in Oracle
+                else if (ctx.datatype().native_datatype_element().VARCHAR2() != null ||
+                        ctx.datatype().native_datatype_element().VARCHAR() != null) {
                     columnEditor
                         .jdbcType(Types.VARCHAR)
                         .type("VARCHAR2");
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDdlParser.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDdlParser.java
@@ -140,9 +140,16 @@ public class OracleDdlParser implements DdlParser {
                     // NUMERIC and DECIMAL types have by default zero scale
                     columnEditor
                         .jdbcType(Types.NUMERIC)
-                        .type("NUMBER")
-                        .length(38)
-                        .scale(0);
+                        .type("NUMBER");
+
+                    if (precisionPart == null) {
+                        columnEditor.length(38)
+                            .scale(0);
+                    }
+                    else {
+                        setPrecision(precisionPart, columnEditor);
+                        setScale(precisionPart, columnEditor);
+                    }
                 }
                 else if (ctx.datatype().native_datatype_element().DATE() != null) {
                     // JDBC driver reports type as timestamp but name DATE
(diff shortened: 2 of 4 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
```diff
@@ -117,6 +123,19 @@ public void readSchema(Tables tables, String databaseCatalog, String schemaNameP
                                     .create()
                                 );
                     }
+                    // NUMBER columns without scale value have it set to -127 instead of null;
```
jpechane: This should be kept in sync with https://github.com/debezium/debezium-incubator/blob/master/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDdlParser.java#L229

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDdlParser.java
```diff
@@ -213,23 +213,34 @@ else if (ctx.datatype().native_datatype_element().FLOAT() != null ||
...
+                    if (precisionPart != null) {
+                        setPrecision(precisionPart, columnEditor);
+                    }
                 }
                 else if (ctx.datatype().native_datatype_element().REAL() != null) {
                     columnEditor
                         .jdbcType(Types.FLOAT)
                         .type("FLOAT")
                         .length(63);
+
```
jpechane: Does `REAL` really have a precision part?
> The REAL datatype is a floating-point number with a binary precision of 63, or 18 decimal.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDdlParser.java
```diff
@@ -298,6 +309,19 @@ private int getVarCharDefaultLength() {
             return 4000;
         }
 
+        private void setPrecision(Precision_partContext precisionPart, ColumnEditor columnEditor) {
```
jpechane: This code (https://github.com/gunnarmorling/debezium-incubator/blob/078a18337cb2bb0fb91faca2b06a8b8869342080/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDdlParser.java#L250) is porbably made obsolete by these two methods

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
```diff
@@ -194,6 +243,85 @@ protected Object convertNumeric(Column column, Field fieldDefn, Object data) {
...
+    protected Object convertTinyInt(Column column, Field fieldDefn, Object data) {
+        if (data == null) {
+            data = fieldDefn.schema().defaultValue();
+        }
+        if (data == null) {
+            if (column.isOptional()) {
+                return null;
+            }
+            else {
+                return BYTE_FALSE;
```
jpechane: In this ocntext the constant should be called `BYTE_ZERO` otherwise it is a little bit misleading.

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/AbstractOracleDatatypesTest.java
```diff
@@ -99,16 +110,27 @@
...
+            new SchemaAndValueField("VAL_DECIMAL", Decimal.builder(6).parameter(PRECISION_PARAMETER_KEY, "10").optional().build(), new BigDecimal("1234.567891")),
+            new SchemaAndValueField("VAL_NUMERIC", Decimal.builder(6).parameter(PRECISION_PARAMETER_KEY, "10").optional().build(), new BigDecimal("1234.567891")),
             new SchemaAndValueField("VAL_NUM_VS", VariableScaleDecimal.builder().optional().build(), VariableScaleDecimal.fromLogical(VariableScaleDecimal.builder().optional().build(), new SpecialValueDecimal(new BigDecimal("77.323"))))
     );
 
     private static final List<SchemaAndValueField> EXPECTED_INT = Arrays.asList(
-            new SchemaAndValueField("VAL_INT", INTEGER_SCHEMA, new BigDecimal("1")),
-            new SchemaAndValueField("VAL_INTEGER", INTEGER_SCHEMA, new BigDecimal("22")),
-            new SchemaAndValueField("VAL_SMALLINT", INTEGER_SCHEMA, new BigDecimal("333"))
+            new SchemaAndValueField("VAL_INT", NUMBER_SCHEMA, new BigDecimal("1")),
```
jpechane: @gunnarmorling Now it feels very strange that we have subset of `NUMERIC` values represented as integer datattypes and integer datatypes defined at the database level are represented as `BigDecimal`. Would not it make it sense to introduce `BigInteger` logical datatype to cover such types? I know that we do the same for extra large types in MySQL but maybe this should be the correct solution from the beginning.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDdlParser.java
gunnarmorling: `REAL` has a fixed precision, i.e. you can't customize it.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDdlParser.java
gunnarmorling: Indeed; I had meant to remove it. Will add it.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
gunnarmorling: Indeed; I had meant to remove it. Will do it.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDdlParser.java
gunnarmorling: Indeed; I had meant to remove it. Will do it.

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/AbstractOracleDatatypesTest.java
gunnarmorling: My thinking is that people generally prefer the built-in KC types over custom ones. In that light, such `BigInteger` type wouldn't be really an advantage over using KC's `DECIMAL` with a scale of 0. Also we'll add a decimal handling mode of `string` which would be an alternative, simpler-to-consume representation. Does that make sense?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDdlParser.java
jpechane: But does not the following code try to work with precision?

gunnarmorling:
For sure worth filing a JIRA issue to spark a discussion and if there's
interest, we could send a PR.

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/AbstractOracleDatatypesTest.java
gunnarmorling: For sure worth filing a JIRA issue to spark a discussion and if there's interest, we could send a PR.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDdlParser.java
gunnarmorling: I'm not quite following. Can you ellaborate a bit? We set the length above (ignoring the fact for now that the value is too large, as 63 is the number of bits, not the number decimal digits, but this shouldn't matter really for now).

gunnarmorling:
@jpechane Pushed two more commits. From my POV it's good, but there's still your pending question regarding `REAL` which isn't quite clear to me.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDdlParser.java
jpechane: See line https://github.com/gunnarmorling/debezium-incubator/blob/0d5f5cfa7eca9899a1645db294e27b01f551fecc/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDdlParser.java#L235 There is no precision part according to the documentation yet you expect that it might arrive

gunnarmorling: Aaaah, gotcha. So those lines L235...238 are superfluous. Thanks! I'll remove them.

gunnarmorling: Btw. another step will be to retrofit this all into the basic parser structure foreseen by Roman. I had hoped to be able to do it for the Alpha1, but it seems it'll not be done by then.

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/AbstractOracleDatatypesTest.java
jpechane: https://issues.apache.org/jira/browse/KAFKA-7199

jpechane:
Depends on https://github.com/debezium/debezium/pull/586

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDdlParser.java
jpechane: I am sorry but my proposal managed to break all other datatypes, see
```
Failed tests: 
  OracleDdlParserTest.shouldParseCreateTable:57 expected:<[1]000> but was:<[4]000>
```

gunnarmorling: Argh, ok. Looking. Sorry for not testing properly locally :( Which reminds me that we really need a PR builder for the incubator repo, including the Oracle connector.

gunnarmorling: Ok, reworked that now, the precision is also applied to `(N)VARCHAR2` columns now. The build is passing with that.

jpechane: Thanks, just a one nitpick - I think `TIMESTAMP` has the same issue?

gunnarmorling: Added a commit for that, too. Any other types where you see it missing?

jpechane:
@gunnarmorling Applied, thanks for the patience!

PR #23, opened 2018-07-25
DBZ-818 Adjusting scale of BigDecimal values to column's scale if needed

gunnarmorling:
https://issues.jboss.org/browse/DBZ-818

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
@@ -235,21 +235,26 @@ public class OracleValueConverters extends JdbcValueConverters {
             }
         }
 
+        // adjust scale to column's scale if the column's scale is larger than the one from
+        // the value (e.g. 4.4444 -> 4.444400)
+        if (data instanceof BigDecimal) {
+            data = withScaleAdjustedIfNeeded(column, (BigDecimal) data);
+        }
+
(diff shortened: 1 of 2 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
```diff
@@ -235,21 +235,26 @@ protected Object convertDecimal(Column column, Field fieldDefn, Object data) {
...
-    @Override
-    protected Object convertNumeric(Column column, Field fieldDefn, Object data) {
-        if (data instanceof NUMBER) {
-            try {
-                data = ((NUMBER)data).bigDecimalValue();
-            }
-            catch (SQLException e) {
-                throw new RuntimeException("Couldn't convert value for column " + column.name(), e);
-            }
+    private BigDecimal withScaleAdjustedIfNeeded(Column column, BigDecimal data) {
```
jpechane: The same code is present in postgres connector - maybe this is something that is common for all connectors https://github.com/debezium/debezium/blob/master/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresValueConverter.java#L346

gunnarmorling:
Yes, it's true. I wasn't quite sure though about the impact of changing it
globally, so I thought I'd stay on the safe side and apply it just here.
Can you open a JIRA issue for it? Thanks!

jpechane:
https://issues.jboss.org/browse/DBZ-825

jpechane:

PR #24, opened 2018-07-25
DBZ-826 Avoid NPE if commit is called before any offset is prepared

gunnarmorling:
https://issues.jboss.org/browse/DBZ-826


jpechane:
@gunnarmorling it works, thanks

PR #30, opened 2018-09-05
DBZ-897 Skip messages redelivered from database after restart

jpechane:
https://issues.jboss.org/browse/DBZ-897


gunnarmorling:
Sweet; applied. Thanks, @jpechane!

PR #36, opened 2018-10-20
DBZ-941 Add "snapshot" snapshot locking mode

grzegorz8:
The entire initial snapshot process is executed in snapshot transaction
isolation level.

PR #38, opened 2018-10-24
DBZ-947 SQL Server snapshot.locking.modes overhaul

grzegorz8:
https://issues.jboss.org/browse/DBZ-947

Exclusive mode: In order to provide a consistent snapshot, exclusive locks
are taken on all monitored tables during entire snapshot process duration.
    
Snapshot mode: The initial snapshot is executed in snapshot transaction
isolation level. This guarantees consistent snapshot as long as DDL
statements are not executed. In addition, neither table locks nor
row-level locks are acquired.
    
Repeatable read mode: The initial snapshot is executed in repeatable read
transaction level. Since phantom reads can occur, it does not fully
guarantee consistency.
    
None mode: Neither table locks nor row-level locks are taken. This way
other transactions are not affected by initial snapshot process.
However, snapshot consistency is not guaranteed. In addition, DDL
statements must not be executed during the snapshot.

### DIFF
--- a/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorConfig.java
+++ b/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorConfig.java
@@ -126,14 +126,7 @@ public class SqlServerConnectorConfig extends HistorizedRelationalDatabaseConnec
          * This mode uses REPEATABLE READ isolation level.  This mode will avoid taking any table locks
          * during the snapshot process.  Since phantom reads can occur, it does not fully guarantee consistency.
          */
-        REPEATABLE_READ("repeatable_read"),
-
-        /**
-         * The snapshot uses READ COMMITTED isolation level.  Neither table locks nor row-level locks are taken.
-         * This way other transactions are not affected by initial snapshot process.  However, snapshot consistency is
-         * not guaranteed.
-         */
-        READ_COMMITTED("read_committed");
+        REPEATABLE_READ("repeatable_read");
 
         private final String value;
 
@@ -212,10 +205,9 @@ public class SqlServerConnectorConfig extends HistorizedRelationalDatabaseConnec
             .withWidth(Width.SHORT)
             .withImportance(Importance.LOW)
             .withDescription("Controls which transaction isolation level is used and how long the connector locks the monitored tables for snapshot execution. "
-                + "The default is '" + SnapshotIsolationMode.REPEATABLE_READ.getValue() + "', which means that the connector does not hold any locks for all monitored tables. "
+                + "The default is '" + SnapshotIsolationMode.REPEATABLE_READ.getValue() + "', which means that the connector does not hold any table locks for all monitored tables. "
                 + "Using a value of '" + SnapshotIsolationMode.EXCLUSIVE.getValue() + "' ensures that the connector holds the exclusive lock (and thus prevents any reads and updates) for all monitored tables. "
-                + "When '" + SnapshotIsolationMode.SNAPSHOT.getValue() + "' is specified, connector runs the initial snapshot in SNAPSHOT isolation level, which guarantees snapshot consistency. In addition, neither table nor row-level locks are held. "
-                + "In '" + SnapshotIsolationMode.READ_COMMITTED.getValue() + "' mode neither table nor row-level locks are acquired, but connector does not guarantee snapshot consistency.");
+                + "When '" + SnapshotIsolationMode.SNAPSHOT.getValue() + "' is specified, connector runs the initial snapshot in SNAPSHOT isolation level, which guarantees snapshot consistency. In addition, neither table nor row-level locks are held.");
 
     /**
      * The set of {@link Field}s defined as part of this configuration.
--- a/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorConfig.java
+++ b/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorConfig.java
@@ -126,7 +126,14 @@ public class SqlServerConnectorConfig extends HistorizedRelationalDatabaseConnec
          * This mode uses REPEATABLE READ isolation level.  This mode will avoid taking any table locks
          * during the snapshot process.  Since phantom reads can occur, it does not fully guarantee consistency.
          */
-        REPEATABLE_READ("repeatable_read");
+        REPEATABLE_READ("repeatable_read"),
+
+        /**
+         * This mode uses READ UNCOMMITTED isolation level.  This mode takes neither table locks nor row-level locks
+         * during the snapshot process.  This way other transactions are not affected by initial snapshot process.
+         * However, snapshot consistency is not guaranteed.
+         */
+        READ_UNCOMMITTED("read_uncommitted");
 
         private final String value;
 
@@ -207,7 +214,8 @@ public class SqlServerConnectorConfig extends HistorizedRelationalDatabaseConnec
             .withDescription("Controls which transaction isolation level is used and how long the connector locks the monitored tables for snapshot execution. "
                 + "The default is '" + SnapshotIsolationMode.REPEATABLE_READ.getValue() + "', which means that the connector does not hold any table locks for all monitored tables. "
                 + "Using a value of '" + SnapshotIsolationMode.EXCLUSIVE.getValue() + "' ensures that the connector holds the exclusive lock (and thus prevents any reads and updates) for all monitored tables. "
-                + "When '" + SnapshotIsolationMode.SNAPSHOT.getValue() + "' is specified, connector runs the initial snapshot in SNAPSHOT isolation level, which guarantees snapshot consistency. In addition, neither table nor row-level locks are held.");
+                + "When '" + SnapshotIsolationMode.SNAPSHOT.getValue() + "' is specified, connector runs the initial snapshot in SNAPSHOT isolation level, which guarantees snapshot consistency. In addition, neither table nor row-level locks are held. "
+                + "In '" + SnapshotIsolationMode.READ_UNCOMMITTED.getValue() + "' mode neither table nor row-level locks are acquired, but connector does not guarantee snapshot consistency.");
 
     /**
      * The set of {@link Field}s defined as part of this configuration.
--- a/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorConfig.java
+++ b/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorConfig.java
@@ -105,9 +105,9 @@ public class SqlServerConnectorConfig extends HistorizedRelationalDatabaseConnec
     }
 
     /**
-     * The set of predefined Snapshot Locking Mode options.
+     * The set of predefined Snapshot Isolation Mode options.
      */
-    public static enum SnapshotLockingMode implements EnumeratedValue {
+    public enum SnapshotIsolationMode implements EnumeratedValue {
 
         /**
          * This mode will block all reads and writes for the entire duration of the snapshot.
@@ -123,14 +123,21 @@ public class SqlServerConnectorConfig extends HistorizedRelationalDatabaseConnec
         SNAPSHOT("snapshot"),
 
         /**
-         * This mode will avoid using ANY table locks during the snapshot process.  This mode can only be used with SnapShotMode
-         * set to schema_only or schema_only_recovery.
+         * This mode uses REPEATABLE READ isolation level.  This mode will avoid taking any table locks
+         * during the snapshot process.  Since phantom reads can occur, it does not fully guarantee consistency.
          */
-        NONE("none");
+        REPEATABLE_READ("repeatable_read"),
+
+        /**
+         * The snapshot uses READ COMMITTED isolation level.  Neither table locks nor row-level locks are taken.
+         * This way other transactions are not affected by initial snapshot process.  However, snapshot consistency is
+         * not guaranteed.
+         */
+        READ_COMMITTED("read_committed");
 
         private final String value;
 
-        private SnapshotLockingMode(String value) {
+        private SnapshotIsolationMode(String value) {
             this.value = value;
         }
 
@@ -145,10 +152,10 @@ public class SqlServerConnectorConfig extends HistorizedRelationalDatabaseConnec
          * @param value the configuration property value; may not be null
          * @return the matching option, or null if no match is found
          */
-        public static SnapshotLockingMode parse(String value) {
+        public static SnapshotIsolationMode parse(String value) {
             if (value == null) return null;
             value = value.trim();
-            for (SnapshotLockingMode option : SnapshotLockingMode.values()) {
+            for (SnapshotIsolationMode option : SnapshotIsolationMode.values()) {
                 if (option.getValue().equalsIgnoreCase(value)) return option;
             }
             return null;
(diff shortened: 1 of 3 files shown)


grzegorz8:
Related pull request: https://github.com/debezium/debezium-incubator/pull/36

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerSnapshotChangeEventSource.java
```diff
@@ -79,31 +78,31 @@ protected SnapshotContext prepare(ChangeEventSourceContext context) throws Excep
...
-            jdbcConnection.connection().setTransactionIsolation(Connection.TRANSACTION_REPEATABLE_READ);
-            ((SqlServerSnapshotContext)snapshotContext).preSchemaSnapshotSavepoint = jdbcConnection.connection().setSavepoint("dbz_schema_snapshot");
+            // No table locks acquired. Read committed isolation level is used to avoid taking shared locks on rows.
+            jdbcConnection.connection().setTransactionIsolation(Connection.TRANSACTION_READ_COMMITTED);
             LOGGER.info("Schema locking was disabled in connector configuration");
-            return;
-        }
-        LOGGER.info("Executing schema locking");
+        } else if (connectorConfig.getSnapshotLockingMode() == SnapshotLockingMode.EXCLUSIVE) {
+            LOGGER.info("Executing schema locking");
```
grzegorz8: Taking locks here is misleading, since it suggests they are taken only for schema snapshot. I believe we should introduce yet another template method.

grzegorz8:
To be honest I have no clue why the build failed.

jpechane:
@grzegorz8 There is a conflict now after https://github.com/debezium/debezium-incubator/pull/36 was applied :-(

grzegorz8:
@jpechane Rebase done.

grzegorz8:
> Hey @grzegorz8, where are we with this one? On first thought I'm having doubts of using READ_COMITTED for snaphotting, as it may lead to concurrent changes being missed. What's the motivation for adding this in addition to REPEATABLE_READ?

I'm waiting for review and suggestions how to structure the code (e.g. do we need to introduce a new template method for taking locks for the entire snapshot, not only for schema snapshot?).

Regarding isolation level, I know that Debezium tries to deliver consistent snapshot, which in fact is feasible with neither READ COMMITTED nor REPEATABLE READ in case of SQL Server. I presented an example for RR here: https://issues.jboss.org/browse/DBZ-947?focusedCommentId=13651038&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13651038
My concern was the fact that "none" mode in current form is misleading. It blocks other transactions to make updates, since in repeatable read isolation level row shared locks are taken until the end of transaction. So I concluded, that if we cannot deliver consistency, we should at least deliver truly "none" locking mode with "eventual consistency".

grzegorz8:
@jpechane @gunnarmorling The PR has been updated according to the suggestions here https://issues.jboss.org/browse/DBZ-947?focusedCommentId=13658596&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13658596

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorConfig.java
```diff
@@ -123,8 +123,14 @@ public static SnapshotMode parse(String value, String defaultValue) {
         SNAPSHOT("snapshot"),
 
         /**
-         * This mode will avoid using ANY table locks during the snapshot process.  This mode can only be used with SnapShotMode
-         * set to schema_only or schema_only_recovery.
+         * This mode will avoid taking any table locks during the snapshot process.  As long as DDL statements are not
+         * executed during snapshot process, data consistency is preserved.
+         */
+        CONSISTENT("consistent"),
```
gunnarmorling: @grzegorz8 How about naming this after the locking mode it uses, i.e. `REPEATABLE_READ`? Makes semantics much clearer, plus, repeatable read isolation still allows for phantom reads, i.e. someone might argue that this doesn't fully guarantee consistency.

gunnarmorling: And should this be the default? It seems much more sensible than `NONE` (and it is more in line with the previous default behavior, where `NONE` was using repeatable read semantics). In fact, is there any reason for keeping `NONE` at this point?

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerSnapshotChangeEventSource.java
```diff
@@ -94,15 +93,20 @@ protected void connectionCreated(SnapshotContext snapshotContext) throws Excepti
     @Override
     protected void lockTablesForSchemaSnapshot(ChangeEventSourceContext sourceContext, SnapshotContext snapshotContext) throws SQLException, InterruptedException {
         if (connectorConfig.getSnapshotLockingMode() == SnapshotLockingMode.NONE) {
+            // No table locks acquired. Read committed isolation level is used to avoid taking shared locks on rows.
+            jdbcConnection.connection().setTransactionIsolation(Connection.TRANSACTION_READ_COMMITTED);
+            LOGGER.info("Schema locking was disabled in connector configuration");
+        } else if (connectorConfig.getSnapshotLockingMode() == SnapshotLockingMode.CONSISTENT) {
```
gunnarmorling: Minor nit: can you move the code after `}` to the next line?

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerSnapshotChangeEventSource.java
gunnarmorling: Yes, that's a good point. Can you file a JIRA issue so we can track it separately?

debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/SnapshotIT.java
```diff
@@ -84,6 +84,11 @@ public void takeSnapshotInSnapshotMode() throws Exception {
         takeSnapshot(SnapshotLockingMode.SNAPSHOT);
     }
 
+    @Test
+    public void takeSnapshotInConsistentMode() throws Exception {
```
gunnarmorling: Seems one for `EXCLUSIVE` is missing, too?

grzegorz8: It's in line 78.

grzegorz8:
@gunnarmorling 
>Do why need to keep NONE at all?
I'm not insisting. I'll remove it.

> Should CONSISTENT be renamed to REPEATABLE_READ and be the new default?
I'll do that.

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorConfig.java
grzegorz8: I agree with both.

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerSnapshotChangeEventSource.java
gunnarmorling: Hum, no, I think I'm taking that back. Before this PR, these locks should indeed only have been held while taking the schema snapshot, by virtue of a savepoint. Which makes me wonder why you removed it here?

grzegorz8:
> Yes, that's a good point. Can you file a JIRA issue so we can track it separately?

@gunnarmorling Which point do you refer to?

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerSnapshotChangeEventSource.java
grzegorz8: done

gunnarmorling:
> I'm not insisting. I'll remove it.

Was just talking to @jpechane a bit, and also re-read your comment above about `REPEATABLE READ` keeping locks. So I can see why someone wanted to use `READ COMITTED`, provided they are aware of the implications. So we can keep it, I'd just suggest to rename it into `READ_COMMITTED` then. And then it'd be a good idea to rename the option and enum into `snapshot.isolation.mode`.

debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/SnapshotIT.java
gunnarmorling: Ah yes. Of course.

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerSnapshotChangeEventSource.java
grzegorz8: Rollback to savepoint makes no sense.
In both SNAPSHOT and R_R mode no locks are taken - no rollback needed.
For EXCLUSIVE, we want to keep locks until the end of transaction - rollback must not be executed.

grzegorz8: I meant that taking exclusive lock in `lockTablesForSchemaSnapshot` is misleading.

gunnarmorling: > For EXCLUSIVE, we want to keep locks until the end of transaction - rollback must not be executed.

I _think_ the original idea was to keep the lock only for the time of taking the schema snapshot. At least that was my motivation for the Oracle connector, which is where @jpechane copied this from. Then the actual data snapshot would be taken using R_R semantics, which indeed seems to be missing still, when looking at the master branch. @jpechane Correct me please, if I'm wrong.

grzegorz8: Yes, on master in EXCLUSIVE data snapshot is taken in database default isolation level.

I suggested to keep the locks during data snapshot as well due to this scenario: https://issues.jboss.org/browse/DBZ-947?focusedCommentId=13651038&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13651038

grzegorz8:
@gunnarmorling I believe I have solved all your suggestions (except `Can you file a JIRA issue so we can track it separately?`). Are there any issues with this PR?

grzegorz8:
Unfortunately, I haven't understood deep enough how read committed works before. It DOES take shared locks, but on very short periods. Deadlocks might occur, though.

```[2018-12-07 18:30:10,367] ERROR Producer failure (io.debezium.pipeline.ErrorHandler:36)
org.apache.kafka.connect.errors.ConnectException: Snapshotting of table XXX.XXX failed
       at io.debezium.relational.HistorizedRelationalSnapshotChangeEventSource.createDataEventsForTable(HistorizedRelationalSnapshotChangeEventSource.java:315)
       at io.debezium.relational.HistorizedRelationalSnapshotChangeEventSource.createDataEvents(HistorizedRelationalSnapshotChangeEventSource.java:259)
       at io.debezium.relational.HistorizedRelationalSnapshotChangeEventSource.execute(HistorizedRelationalSnapshotChangeEventSource.java:132)
       at io.debezium.pipeline.ChangeEventSourceCoordinator.lambda$start$0(ChangeEventSourceCoordinator.java:77)
       at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
       at java.util.concurrent.FutureTask.run(FutureTask.java:266)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
       at java.lang.Thread.run(Thread.java:745)
Caused by: com.microsoft.sqlserver.jdbc.SQLServerException: Transaction (Process ID 2316) was deadlocked on lock resources with another process and has been chosen as the deadlock victim. Rerun the transaction.
       at com.microsoft.sqlserver.jdbc.SQLServerException.makeFromDatabaseError(SQLServerException.java:259)
       at com.microsoft.sqlserver.jdbc.SQLServerResultSet$FetchBuffer.nextRow(SQLServerResultSet.java:6388)
       at com.microsoft.sqlserver.jdbc.SQLServerResultSet.fetchBufferNext(SQLServerResultSet.java:1647)
       at com.microsoft.sqlserver.jdbc.SQLServerResultSet.next(SQLServerResultSet.java:988)
       at io.debezium.relational.HistorizedRelationalSnapshotChangeEventSource.createDataEventsForTable(HistorizedRelationalSnapshotChangeEventSource.java:287)
       ... 8 more```
I need to rethink usefulness of `none` mode :(

grzegorz8:
@jpechane In this case I was fetching one huge table only (it failed after ~4 hours). I've read that while reading table in read committed a read lock or either individual row or a page is acquired. I'm trying now to track down the exact scenario.

gunnarmorling:
Hey @grzegorz8 and @jpechane, it seems this one still needs some more work? Anything specific we should discuss?

jpechane:
@gunnarmorling Yeah, I'd still give it a couple of more rounds. If the locks are on raw/page level then even on table level if a concurrent process will read two rows in different order than Debezium we can end in a deadlock.

gunnarmorling:
Ok, thanks for the heads-up. I'll assign the issue to 0.9-next then.

grzegorz8:
@gunnarmorling I doubt we can do anything about deadlocks. We can either decide to get rid of `read_committed` support or we can leave it and document it properly.

If we leave it, we should document tradeoffs between consistency and locking of each mode and mention that the best choice is SNAPSHOT isolation level and if it's not an option for a user, the user should accept disadvantages of other isolation level modes.

jpechane:
@grzegorz8 @gunnarmorling Based up the current observations I think dropping `read_committed` is fine, I'd still provide `none` for edge cases. Naming is fine for me.

grzegorz8:
@jpechane How should `none` look like? I believe `repeatable_read` has the same semantics as the old `none`.

grzegorz8:
@jpechane @gunnarmorling How about this PR. If the overall idea of the PR is not good for you, let me create separate tasks (and PRs) for bugs fixes here:
a) rollback to a savepoint was removed. In exclusive mode we need to keep lock until the end of transaction. For snapshot and repeatable_read rollback to a savepoint has no effect in fact.
b) Once the snapshot is finished, the default isolation level (read committed) is set.

jpechane:
@grzegorz8 Nobody is saying that the overall idea is not good. I understand this takes a little bit long already but as you've pointed before - not everything works as we all expected. To speed things up how baout having a chat or hangout session where we agree upon final solution? I think it is better than pushing back and forth. @gunnarmorling Do you agree?

grzegorz8:
@jpechane I understand this is rather low-priority task if we consider other improvements, bugs or upcoming release candidate.
A quick chat is fine for me. We can schedule it at your convenience, once you have more free time :)

gunnarmorling:
Yes, good idea with having a chat, so to agree on the way forward. And sorry for the delay, there's just (too) much going on. Would Friday morning, e.g. 10:00 a.m. CET, work for the two of you?

jpechane:
@gunnarmorling Fine for me!

grzegorz8:
Could it be 10:30, please?

gunnarmorling:
Yes, of course. I'll send an invite with a Google Hangouts link.

grzegorz8:
@gunnarmorling Thanks!

gunnarmorling:
Sent an invite to your Google address.

gunnarmorling:
@grzegorz8 Can you close this PR and open a new one against the main repository (after tomorrow's call, I reckon)? In the course of moving the SQL Server connector out of "Incubating" state, we've moved the connector over there. Thanks!

PR #39, opened 2018-10-27
DBZ-953 Ability to configure "decimal.handling.mode" for sql server Connector

pradeepmvn:
Ability to Configure decimal.handling.mode similar to mysql connector

### DIFF
--- a/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorConfig.java
+++ b/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorConfig.java
@@ -17,6 +17,7 @@ import io.debezium.config.Field;
 import io.debezium.document.Document;
 import io.debezium.heartbeat.Heartbeat;
 import io.debezium.jdbc.JdbcConfiguration;
+import io.debezium.jdbc.JdbcValueConverters.DecimalMode;
 import io.debezium.relational.HistorizedRelationalDatabaseConnectorConfig;
 import io.debezium.relational.RelationalDatabaseConnectorConfig;
 import io.debezium.relational.TableId;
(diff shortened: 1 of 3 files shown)


jpechane:
@pradeepmvn Thanks for the contribution! Could you please formatting and use spaces instead of tabs? It is the reason for build failure.

jpechane:
Generally the code looks good but there are many changes related just to formatting. Would it be possible to keep the code changes mostly to the functionality change? Thanks a lot

pradeepmvn:
@jpechane I think you are refering to SqlServerConnectorConfig.java. Yeah true it shows many changes. Let me update that just to have a clear dif..

jpechane:
@pradeepmvn It is much better, thanks.
Could you please fix the Checkstyle errors too?
>/home/travis/build/debezium/debezium-incubator/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerValueConverters.java:45:5: File contains tab characters (this is the first instance).
/home/travis/build/debezium/debezium-incubator/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorConfig.java:206:1: File contains tab characters (this is the first instance).
/home/travis/build/debezium/debezium-incubator/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerDatabaseSchema.java:42:1: File contains tab characters (this is the first instance).

Also would it be possible to add tests to verify that the mode change really reflect the encoding?
And the last one - would it be possible tpo introduce `decimalHandlingMode` method on `SqlServerConnectorConfig` so the caller does not need to do all the parsing and casting?

pradeepmvn:
@jpechane Fixed them
- Fixed checkstyle Errors
-Added test cases for decimal mode property
- Added getDecimalModeConfig to move parsing away from caller functions
Hope this helps.

Also, any info available on when the SQL Server connector be released form incubation?

gunnarmorling:
Nice contribution, @pradeepmvn. Thanks a lot!

There's one thing I'll change before applying, and that is to pull up `DecimalHandlingMode` into the `RelationalDatabaseConnectorConfig` base class. That way we avoid some duplication across connectors.

gunnarmorling:
@pradeepmvn In the meantime, could you already prepare a PR for adding the new option to the [SQL Server connector docs](https://github.com/debezium/debezium.github.io/blob/develop/docs/connectors/sqlserver.asciidoc)? Thanks!

gunnarmorling:
> @pradeepmvn In the meantime, could you already prepare a PR for adding the new option to the SQL Server connector docs? Thanks!

Ha, never mind, just saw you already did so. So all is good; thanks again!

pradeepmvn:
> Nice contribution, @pradeepmvn. Thanks a lot!
> 
> There's one thing I'll change before applying, and that is to pull up `DecimalHandlingMode` into the `RelationalDatabaseConnectorConfig` base class. That way we avoid some duplication across connectors.

@gunnarmorling 
Yes true. I also think existing connectors - Postgres, MYSQL should leverage  a common ```DecimalHandlingMode``` enum. There is a decimal handling mode enum defined  under ```JdbcValueConverters``` that can be leveraged too.

PR #40, opened 2018-10-28
DBZ-954 Oracle Connector doesn't support Oracle 11g

wangzheng422:
The original dbz can't run with oracle 11g, so add 2 parameters:
database.tablename.case.insensitive, default is false, set to true for oracle 11g.
database.position.version, default is v2, set to v1 for oracle 11g.


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -80,6 +81,19 @@
...
                     + "'initial_schema_only' to specify the connector should run a snapshot of the schema when no offsets are available for the logical server name. ");
 
+    public static final Field TABLENAME_CASE_INSENSITIVE = Field.create("database.tablename.case.insensitive")
+        .withDisplayName("Oracle table name case insensitve mode")
+        .withType(Type.BOOLEAN)
+        .withDefault(false)
+        .withImportance(Importance.LOW)
+        .withDescription("Oracle table node case sensitve true or false, use true mode when you need to work with oracle 11g.");
+    
+    public static final Field POS_VERSION = Field.create("database.position.version")
```
jpechane: Maybe we could merge both new fields into a one - called `OracleVersion`. Thet would be enum  that will be configured with values `11g` and `12c` and will provide setting for table name insensitive handling and pos version?

wangzheng422:
I just reformat the code. Later I will squash the commits to another branch and make pull request using the new branch. Now I will think about how to get rid of these two parameters.

gunnarmorling:
Hey @wangzheng422, any update on this one? Would be cool to squash everything into a single commit. And can you confirm that this makes the connector work with Oracle 11 for you?

wangzheng422:
This works with oracle 11g, here is my test case readme:
https://github.com/wangzheng422/docker_env/blob/master/a6test/kafka/dbz-oracle/11.2.0.4/README.md

Here is the setup shell to prepare the oracle env:
https://github.com/wangzheng422/docker_env/blob/master/a6test/kafka/dbz-oracle/11.2.0.4/setup.sh

But I find out 2 bugs around the fix:
1. the table white list must be lower case, if not, there will be dump.
2. in some environment (to me, it is production system), the xstream will stop to receive LCR event, or not process the event, it seems the thread blocks.

I am investigating on these bugs right now, and will give fixes on them.

gunnarmorling:
Hey @wangzheng422, do you plan to send a new PR replacing this one?

wangzheng422:
Yes, use new PR instead.

https://github.com/debezium/debezium-incubator/pull/46

wangzheng422:
Yes, use new PR instead.
https://github.com/debezium/debezium-incubator/pull/46

PR #42, opened 2018-11-06
DBZ-953 Support for "decimal.handling.mode" option for SQL Server

gunnarmorling:
https://issues.jboss.org/browse/DBZ-953

Supersedes #39.

Requires core PR https://github.com/debezium/debezium/pull/666.


jpechane:
@gunnarmorling Applied, thanks! I've just expanded tests with schema assertions.

PR #45, opened 2018-11-16
DBZ-776 Removing redundant method;

gunnarmorling:
Adjusting test for database property pass-through, as the connection isn't established any longer synchronously within start().


gunnarmorling:
Thought it would come from that, but indeed it's older actually. But seems not worth creating its own issue IMO.

jpechane:

PR #47, opened 2018-11-22
DBZ-1002 Avoiding NPEs when connector failed to start

gunnarmorling:
https://issues.jboss.org/browse/DBZ-1002


jpechane:

PR #50, opened 2018-11-30
DBZ-1017 Record schema evolution into history correctly

jpechane:
https://issues.jboss.org/browse/DBZ-1017

### DIFF
--- a/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerDatabaseSchema.java
+++ b/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerDatabaseSchema.java
@@ -48,7 +48,7 @@ public class SqlServerDatabaseSchema extends HistorizedRelationalDatabaseSchema 
         tables().overwriteTable(table);
 
         TableChanges tableChanges = null;
-        if (schemaChange.getType() == SchemaChangeEventType.CREATE && schemaChange.isFromSnapshot()) {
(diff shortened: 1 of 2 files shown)


debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerDatabaseSchema.java
```diff
@@ -48,7 +48,7 @@ public void applySchemaChange(SchemaChangeEvent schemaChange) {
         tables().overwriteTable(table);
 
         TableChanges tableChanges = null;
-        if (schemaChange.getType() == SchemaChangeEventType.CREATE && schemaChange.isFromSnapshot()) {
+        if (schemaChange.getType() == SchemaChangeEventType.CREATE) {
```
gunnarmorling: @jpechane, why is it that only CREATE statements are tracked?

jpechane: @gunnarmorling The fact is that from the implementaiton POV there is no difference between `CREATE` and `ALTER` so we are using a single event.

gunnarmorling: Ok; I have to admit this still is a bit blurry to me. Would be nice to get some clarity in the context of [DBZ-993](https://issues.jboss.org/browse/DBZ-993).

gunnarmorling:
Rebased and applied. Thanks, @jpechane!

PR #51, opened 2018-12-05
DBZ-978 Added last event and captured tables metrics

jpechane:
Requires https://github.com/debezium/debezium/pull/704

### DIFF
--- a/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerTaskContext.java
+++ b/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerTaskContext.java
@@ -16,6 +16,6 @@ import io.debezium.connector.common.CdcSourceTaskContext;
 public class SqlServerTaskContext extends CdcSourceTaskContext {
 
     public SqlServerTaskContext(SqlServerConnectorConfig config, SqlServerDatabaseSchema schema) {
-        super("SQLServer", config.getLogicalName(), schema::tableIds);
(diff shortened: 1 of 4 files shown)


debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerTaskContext.java
```diff
@@ -15,7 +15,7 @@
  */
 public class SqlServerTaskContext extends CdcSourceTaskContext {
 
-    public SqlServerTaskContext(SqlServerConnectorConfig config) {
-        super("SQL Server", config.getLogicalName());
+    public SqlServerTaskContext(SqlServerConnectorConfig config, SqlServerDatabaseSchema schema) {
+        super("SQLServer", config.getLogicalName(), schema::tableIds);
```
gunnarmorling: Is this the correct spelling? I find "SQL Server" (two words) more often on the web.

jpechane: @gunnarmorling I originally used it too but it is not practical when it is used in JMX names and dashboard queries

gunnarmorling: Ok, I get the reasoning. I'd suggest to rename it to "SQL_Server" then. I'll do that change and apply.

gunnarmorling:
Rebased and applied. Thanks, @jpechane!

PR #61, opened 2019-01-10
DBZ-1064 Set default value only for non-nullable fields

jpechane:
Main issue - https://github.com/debezium/debezium/pull/723


gunnarmorling:
@jpechane, this should still be updated to use the new base method where applicable.

jpechane:
@gunnarmorling Yes, I waited for the main PR to be merged to make sure there is no need to rewirte

gunnarmorling:
Makes sense! Just merged that one :)

jpechane:
Requires https://github.com/debezium/debezium/pull/735 to pass failing test

gunnarmorling:
Applied. Thanks, @jpechane!

PR #62, opened 2019-01-10
DBZ-1069 Snapshot is completed when offset is recovered

jpechane:


jpechane:
https://issues.jboss.org/browse/DBZ-1069

debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/SqlServerConnectorIT.java
```diff
@@ -256,6 +257,66 @@ public void streamChangesWhileStopped() throws Exception {
...
+            );
+            connection.execute(
+                    "INSERT INTO tableb VALUES(" + id + ", 'b')"
+            );
+        }
+
+        start(SqlServerConnector.class, config);
+        assertConnectorIsRunning();
+        records = consumeRecordsByTopic(1 + RECORDS_PER_TABLE * TABLES).allRecordsInOrder();
+        records = records.subList(1, records.size());
```
grzegorz8: It seems to be wrong. If you have a look at records, they have ids: 10, 11, 12, ... Should be 100, 101.

grzegorz8: This means that the same records have been fetched twice - during initial load and streaming.

jpechane: You are right, but this is a different issue - I need to think about it more. The last snapshot record should probably contan the LSN from which we should start reading in streaming phase.

grzegorz8:
@jpechane Looks good now :+1:

gunnarmorling:
@grzegorz8, does the PR resolve the original issue for you? I'm a bit unsure given your last comment on the JIRA issue.

grzegorz8:
@gunnarmorling Yes, it does :)

gunnarmorling:
Ok, thanks!

PR #64, opened 2019-01-15
DBZ-1051 Handle object and db name letter case

jpechane:
https://issues.jboss.org/browse/DBZ-1051

### DIFF
--- a/debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/CaseSensitivenessIT.java
+++ b/debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/CaseSensitivenessIT.java
@@ -86,7 +86,7 @@ public class CaseSensitivenessIT extends AbstractConnectorTest {
             (Struct)((Struct)record.value()).get("after"),
             SchemaBuilder.struct()
                 .optional()
-                .name("server1.testDB.dbo.MyTableOne.Value")
+                .name("server1.dbo.MyTableOne.Value")
                 .field("Id", Schema.INT32_SCHEMA)
                 .field("ColA", Schema.OPTIONAL_STRING_SCHEMA)
                 .build()
@@ -94,7 +94,7 @@ public class CaseSensitivenessIT extends AbstractConnectorTest {
         assertSchemaMatchesStruct(
             (Struct)record.key(),
             SchemaBuilder.struct()
-                .name("server1.testDB.dbo.MyTableOne.Key")
+                .name("server1.dbo.MyTableOne.Key")
                 .field("Id", Schema.INT32_SCHEMA)
                 .build()
         );
@@ -108,7 +108,7 @@ public class CaseSensitivenessIT extends AbstractConnectorTest {
             (Struct)((Struct)record.value()).get("after"),
             SchemaBuilder.struct()
                 .optional()
-                .name("server1.testDB.dbo.MyTableOne.Value")
+                .name("server1.dbo.MyTableOne.Value")
                 .field("Id", Schema.INT32_SCHEMA)
                 .field("ColA", Schema.OPTIONAL_STRING_SCHEMA)
                 .build()
@@ -116,7 +116,7 @@ public class CaseSensitivenessIT extends AbstractConnectorTest {
         assertSchemaMatchesStruct(
             (Struct)record.key(),
             SchemaBuilder.struct()
-                .name("server1.testDB.dbo.MyTableOne.Key")
+                .name("server1.dbo.MyTableOne.Key")
                 .field("Id", Schema.INT32_SCHEMA)
                 .build()
         );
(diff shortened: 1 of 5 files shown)


debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/CaseSensitivenessIT.java
```diff
@@ -0,0 +1,148 @@
...
+        connection.execute(
+                "CREATE TABLE MyTableOne (Id int primary key, ColA varchar(30))",
+                "INSERT INTO MyTableOne VALUES(1, 'a')"
+        );
+        TestHelper.enableTableCdc(connection, "MyTableOne");
+        testDatabase();
+    }
+
+    @Test
+    public void caseSensitiveDatabase() throws Exception {
```
gunnarmorling: How does this test the originally reported issue?

debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/CaseSensitivenessIT.java
```diff
@@ -0,0 +1,148 @@
...
+        assertConnectorIsRunning();
+
+        SourceRecords records = consumeRecordsByTopic(1);
+        Assertions.assertThat(records.recordsForTopic("server1.dbo.MyTableOne")).hasSize(1);
+        SourceRecord record = records.recordsForTopic("server1.dbo.MyTableOne").get(0);
+        assertSchemaMatchesStruct(
+            (Struct)((Struct)record.value()).get("after"),
+            SchemaBuilder.struct()
+                .optional()
+                .name("server1.testDB.dbo.MyTableOne.Value")
```
gunnarmorling: Not directly related to this PR, but this schema name makes me wonder. What is that "dbo" part in there? Why isn't it "server1.testDB.MyTableOne.Value", i.e. as it'd be for the other connector? In particular, as the "testDB" part is missing in the topic name above. This seems like something we should straighten out?

debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/CaseSensitivenessIT.java
jpechane: The original issue is verified by this change - https://github.com/debezium/debezium-incubator/pull/64/files#diff-b306088505d9239e03c8adb3da8488deR27

debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/CaseSensitivenessIT.java
jpechane: `dbo` is schema name. Connector can connect to only single database, hence I follow the same approach as https://debezium.io/docs/connectors/postgresql/#topic-names so the database name is droppend and replaced with the logical name.

debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/CaseSensitivenessIT.java
gunnarmorling: Uff, ok, that's subtle :)

debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/CaseSensitivenessIT.java
gunnarmorling: > so the database name is droppend and replaced with the logical name.

That makes sense, but it's not what's currently done, is it? I.e. it should be `server1.dbo.MyTableOne.Value` instead of `server1.testDB.dbo.MyTableOne.Value`, or am I missing something?

gunnarmorling: Cool; so btw. that has been my general conclusion on this subject: the DBZ connectors should always only connect to a single database in case a specific DB host can have multiple ones. That database is then represented by the "logical name" in topic and message schema names.

jpechane: This one is a little bit tricky - we are creating envelop schema name from topic name but key/value schema name from table id. Table id is triplet of catalog, schema and table. MySQL has schema set to null, PostgreSQL and Oracle has catalog set to null but SQL Server is really hard-working guy and sets both catalog and and schema which leads to that unusual schema name.

jpechane: @gunnarmorling Updated, now depends on https://github.com/debezium/debezium/pull/739

gunnarmorling: > SQL Server is really hard-working guy

😄 

> Updated

Nice, thanks!

PR #68, opened 2019-01-23
DBZ-1084 Removing SQL Server Connector as it's in the main repo now

gunnarmorling:
@jpechane, this removes the SQL Server connector from the incubator repo.

https://issues.jboss.org/projects/DBZ/issues/DBZ-1084


gunnarmorling:
@jpechane, here's the removal of the SQL Server connector in the incubator repo.

PR #74, opened 2019-02-28
DBZ-1166 NullPointerException When there are multiple tables in different schemas in the whitelist

elgca:
Solve the problem of NullPointerException when there are multiple schema


jpechane:
@elgca Hi, thanks for the PR, it is very appreciated! Could you please create a Jira issue with the problem description and modify the commit message to be in format `DBZ-xxx commit message`?

elgca:
> @elgca Hi, thanks for the PR, it is very appreciated! Could you please create a Jira issue with the problem description and modify the commit message to be in format `DBZ-xxx commit message`?

issues created.

jpechane:
@elgca Thanks, applied! I've just reworded the commit message and added a test

PR #77, opened 2019-03-22
DBZ-137 Convert Oracle DDL Antlr parser to mimic MySQL

jpechane:


gunnarmorling:
Applied. Thank you so much, @nyiny123! So we probably should talk about the next steps, in particular using LogMiner. I remember you made some progress on that end, right (sorry btw. for all the delays with replying, I was/am just swamped with so many things). Should we do a call to talk about it? Would Friday work for you?

nyiny123:
Hi, Gunnar
I understand you are overwhelmed.
This week my company holds a conference and I might be distracted any time. I’m planning to submit DML parser PR tomorrow, when you take a look, we may have better view of the next steps. We may have a meeting next week.
Andrey

PR #81, opened 2019-04-11
DBZ-212 Refactored LOGICAL_NAME to Parent Abstract Class

krizhan:
Related to https://github.com/debezium/debezium/pull/809


gunnarmorling:
Thanks, @krizhan!

PR #88, opened 2019-05-27
DBZ-1235 Oracle converted to use source info maker

jpechane:
Depends on https://github.com/debezium/debezium/pull/886


debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/SourceInfoTest.java
```diff
@@ -18,8 +21,15 @@
 
     @Before
     public void beforeEach() {
```
gunnarmorling: How about adding a test that asserts the entire source info as you've added it for the other connectors too?

jpechane: @gunnarmorling Added

gunnarmorling:
Applied; thanks, @jpechane!

PR #95, opened 2019-06-13
DBZ-1317 Removing redundant method

gunnarmorling:
@jpechane A small left-over. Technically, the issue has already been resolved in Beta1, but I think it's ok to add this one here without a new issue.


jpechane:

PR #99, opened 2019-07-01
DBZ-1356 Metrics for schema history

jpechane:
https://issues.jboss.org/browse/DBZ-1356

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/Module.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/Module.java
@@ -25,4 +25,8 @@ public final class Module {
     public static String name() {
         return "oracle";
     }
+
+    public static String connectorName() {
+        return "Oracle";
+    }
 }
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/Module.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/Module.java
@@ -22,11 +22,17 @@ public final class Module {
         return INFO.getProperty("version");
     }
 
+    /**
+     * @return symbolic name of the connector plugin
+     */
     public static String name() {
         return "oracle";
     }
 
-    public static String connectorName() {
+    /**
+     * @return context name used in log MDC and JMX metrics
+     */
+    public static String contextName() {
         return "Oracle";
     }
 }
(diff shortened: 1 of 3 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/Module.java
```diff
@@ -25,4 +25,8 @@ public static String version() {
     public static String name() {
         return "oracle";
     }
+
+    public static String connectorName() {
```
gunnarmorling: @jpechane so what's the different usages of `name()` vs. `connectorName()`. Wondering whether we can consolidate those two actually?

jpechane: @gunnarmorling One is used in JMX metrics and the other in sourceinfo. Unfrotunately some connectors have different vale for these two so for backward compatibility I'd keep them separated.

gunnarmorling: It seems though `name()` could be derived from `connectorName()` by lower-casing and removing the "_"? I.e. `name()` could look like so in the base class:

```
public static String name() {
        return connectorName().toLowerCase().replaceAll("_", "");
}
```

WDYT?

And the two methods should have JavaDoc on their purpose in the base class, too.

jpechane: @gunnarmorling I thought about it too but check postgres module file

jpechane: @gunnarmorling And do not forget that static method calls have early not late binding so inheritance does not work.

jpechane: @gunnarmorling Thinking about it more, postgresql metrics are not exposed yet. The JMX metrics names are lower-cased anyway so the only porblem left is SQLServer that has metrics exposed with `_`.
The other impact is just the logging context name in log files which I believe we cann afford to change

gunnarmorling: On SQL Server, see above: couldn't we simply exclude the "_"? It seems a reasonably generic thing to do.

PR #100, opened 2019-07-03
bug fix for DBZ-1354

dreamcatchernick:
Directly copy decompiled source code from original XStreamUtility, fix bug for converting rawPosition(byte[]) to long value from XStream LCR.getPosition()

If rawPosition is a large number, the original method will return a negative value which makes nonsense.Because according to oracle doc, scn should always be a positive value.

An incorrect scn number will impact the debezium oralce connector behavior. See more details in DBZ-1354


gunnarmorling:
@dreamcatchernick, thanks, but we cannot apply any code derived from decompling proprietary code! The only acceptable way would be a clean-room implemention of the affected method, solely based on the publicly documented behavior. I'm going to close this one.

dreamcatchernick:
ok i will remove the code. Sorry about that

PR #101, opened 2019-07-04
bug fix for DBZ-1354

dreamcatchernick:
Fix bug for converting rawPosition(byte[]) to long value from XStream LCR getPosition()

If rawPosition is a large number, the original method will return a negative value which makes nonsense.Because according to oracle doc, scn should always be a positive value.

An incorrect scn number will impact the debezium oracle connector behavior. See more details in DBZ-1354


jpechane:
@gunnarmorling @dreamcatchernick I think that the best legally sound solution is to use
https://docs.oracle.com/cd/E18283_01/appdev.112/e16760/t_lcr.htm#BGBFJGEB
https://docs.oracle.com/cd/E18283_01/appdev.112/e16760/t_lcr.htm#BGBDDBDJ

dreamcatchernick:
> @gunnarmorling @dreamcatchernick I think that the best legally sound solution is to use
> https://docs.oracle.com/cd/E18283_01/appdev.112/e16760/t_lcr.htm#BGBFJGEB
> https://docs.oracle.com/cd/E18283_01/appdev.112/e16760/t_lcr.htm#BGBDDBDJ
we already use getSCNFromPosition() , however after i test the method, i think there is a bug in that method. so decide re-write a correct one

jpechane:
@dreamcatchernick - Well my idea was to use JDBC call to invoke pointed stored procedures to get the necessary values - thus we will not need to rewirte the Java method. Does the stored procedure return the correct value or is broken too?

gunnarmorling:
@dreamcatchernick, can you please check whether calling this stored procedure is a viable solution? This is the only way I can see us handling this really (apart from reporting the issue upstream of course and having people use a version of the driver which contains the fix).

dreamcatchernick:
hi @jpechane @gunnarmorling 
I did quick try to use the store procedure. unfortunately i couldn't manage to invoke the procedure via ojdbc. Below is my code snippet to call the procedure
````
            Connection connection = DriverManager.getConnection("jdbc:oracle:thin:@10.177.83.9:1521/ycdb", "xstrmadmin" , "xsa");

            String sql = "{call LCR$_ROW_RECORD.GET_SCN_FROM_POSITION(?)}";

            Statement statement = connection.createStatement();

            ResultSet resultSet = statement.executeQuery("select CURRENT_SCN from V$DATABASE");
            if (!resultSet.next()) {
                throw new IllegalStateException("Couldn't get SCN");
            }

            long current_scn = resultSet.getLong(1);
            byte[] rawPostion = XStreamUtility.convertSCNToPosition(new NUMBER(current_scn), XStreamUtility.POS_VERSION_V1);

            CallableStatement callableStatement = connection.prepareCall(sql);
            callableStatement.setBytes(1, rawPostion);
            //callableStatement.registerOutParameter(2, OracleTypes.NUMBER);
            try{
                boolean execute = callableStatement.execute();
            } catch ( Exception e) {
                System.out.println(e);
            }

````
it throws exception 'PLS-00201: identifier 'GET_SCN_FROM_POSITION' must be declared'

the user "xstrmadmin" should have enough privileges. 

I am not an oracle pro. Any assistant would be  appreciated.

However i am not quite agree to use that store procedure. Two reasons:

1.  this would be a remote call, there will be additional effort. This approach actually changes the original design.   
2. the user to invoke the store procedure need to have certain privileges which brings more config on oracle. Specially for enterprise dev and db are two separate departments,  more communication will be added.

Thanks
Nick

dreamcatchernick:
@gunnarmorling @jpechane hi i manage to get that FUNCTION works
here is the query string
````
String sql = "{? = call SYS.LCR$_ROW_RECORD.GET_SCN_FROM_POSITION(?)}";
````
i confirms it works on 11g which returns me correct scn number.

However , the original statement "XStreamUtility.getSCNFromPosition(rawPosition).longValue();"   is in LcrPosition constructor , if we replace it to that query this will have huge impact on the performance.

gunnarmorling:
Can you quantify the performance impact?

dreamcatchernick:
@gunnarmorling according to the code. Every time receives LCR from XStream it will try to convert a raw position(byte[]) to long value. 
if we replace to use ORACLE FUNCTION, for each single LCR it will invoke the FUNCTION which would be a  massive net/io cost

gunnarmorling:
Sure, but what does it mean in terms of actually increased latency of events? I understand it's not ideal, but unfortunately I cannot see any viable alternative at this point.

dreamcatchernick:
@gunnarmorling  the solution at this stage i can think of 
1. long term : report to oracle let them to fix XStreamUtility(this will totally depend on them)

2. medium term: Replace use the "ORACLE FUNCTION" . For each LCR leave it as byte array convert to long value until it is needed. However this will bring quite amount of refactor.

3. short term: just replace to use that ORACLE FUNCTION which means the more data receives from XStream the more latency and net/io will increased.

4. quickest way: Do nothing -_- (Oracle 11g will always has that bug. 12+should be fine)

please add more solution if you can think of
cheers

jpechane:
@dreamcatchernick Hi, thanks for the effort. I've checked the code and IMHO SCN is almost always needed just after the constructor is called so it is probably not necessary to do such optimization.
If you feel it is still needed then I propose just to make the initialization lazy in `getScn` method. No other code needs to be touched.
Ideally the `LcrPosition` will support using both call to `XStreamUtility` and the `ORACLE FUNCTION` and it will be configurable which way it will be used - depending on Oracle version.

dreamcatchernick:
@jpechane  why is that scn have to convert to long value? can we just leave as raw byte array?

jpechane:
@dreamcatchernick Hi, you are almost right. We are also using `LcrPosition` and the conversion is done only i cases where the position is not available. Unfortunately the problem lies with snapshotting. When we execute the snapshot the only information we have available is SCN at time of snapshot. We then convert the scn to byte[] so we can rewind the position in stream to changes that happened after snapshot. But taking this into consideration the conversion, The scn is then used only for informational purposes.
So maybe the correct solution is this
 * support conversions both via Oracle code and via stored procedure
 * use store procedure converions only at the beginning of `io.debezium.connector.oracle.OracleStreamingChangeEventSource.execute(ChangeEventSourceContext)` - this is only place where it is really needed to have it right
 * in all other cases keep it as is

This means that the expensive call is done only once why connector is starting streaming. WDYT?

gunnarmorling:
I'm going to close this one, as it's out of date, and we're focusing our efforts on the LogMiner implementation at this point. Thanks for your efforts in any case.

PR #107, opened 2019-09-23
DBZ-1044 Sanitize column/topic names to always be valid AVRO names

Naros:
https://issues.jboss.org/browse/DBZ-1044


gunnarmorling:
Thanks, @Naros!

PR #109, opened 2019-09-30
[DBZ-1408] Refactor SourceInfo to extend AbstractSourceInfo

bingqinzhou:
https://issues.jboss.org/projects/DBZ/issues/DBZ-1408

This PR includes the following major changes:
- Refactor CassandraConnectorConfig to extend CommonConnectorConfig.
- Refactor SourceInfo to extend AbstractSourceInfo.


gunnarmorling:
Hey @criccomini and @jgao54, any feedback from your side on this one?

@bingqinzhou, could you rebase it to the latest master, as there's a conflict. Thanks!

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
```diff
@@ -373,70 +357,72 @@ public CommitLogTransfer getCommitLogTransfer() {
...
     }
 
+    /**
+     * Whether deletion events should have a subsequent tombstone event (true) or not (false).
+     * It's important to note that in Cassandra, two events with the same key may be updating
+     * different columns of a given table. So this could potentially result in records being lost
+     * during compaction if they haven't been consumed by the consumer yet. In other words, do NOT
+     * set this to true if you have kafka compaction turned on.
+     */
     public boolean tombstonesOnDelete() {
```
gunnarmorling: That's already handled by `CommonConnectorConfig` (see `isEmitTombstoneOnDelete()` and related code).

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
```diff
@@ -461,16 +447,31 @@ public Converter getValueConverter() throws CassandraConnectorConfigException {
...
     }
+
+    @Override
+    public String getContextName() {
+        return Module.contextName();
+    }
+
+    @Override
+    protected SourceInfoStructMaker<? extends AbstractSourceInfo> getSourceInfoStructMaker(Version version) {
+        switch (version) {
```
gunnarmorling: This shouldn't be needed, or do we actually want to maintain two different versions of the source struct? I don't think so, given the connector's "incubating" state. The notion of the "legacy" struct maker exists for the other connectors, as we did some technically non-backwards compatible changes to `source` in 0.10 (for the sake of general consistency and stream-lining, this might btw. relevant for you when upgrading your MySQL connectors) and we wanted to have a transitioning period where the old one still is available, too. This shouldn't be needed for Cassandra.

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/LegacyV1CassandraSourceInfoStructMaker.java
```diff
@@ -0,0 +1,50 @@
...
+package io.debezium.connector.cassandra;
+
+import io.debezium.config.CommonConnectorConfig;
+import io.debezium.connector.LegacyV1AbstractSourceInfoStructMaker;
+import io.debezium.connector.cassandra.transforms.CassandraTypeKafkaSchemaBuilders;
+import io.debezium.util.SchemaNameAdjuster;
+import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.data.Struct;
+
+public class LegacyV1CassandraSourceInfoStructMaker extends LegacyV1AbstractSourceInfoStructMaker<SourceInfo> {
```
gunnarmorling: See my comment above; shouldn't be needed.

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/SourceInfo.java
```diff
@@ -108,4 +84,19 @@ public String toString() {
...
+    }
+
+    @Override
+    protected SnapshotRecord snapshot() {
+        return snapshot ? SnapshotRecord.TRUE : SnapshotRecord.FALSE;
+    }
+
+    @Override
+    protected String database() {
+        return "NULL";
```
gunnarmorling: Could you add a clarifying comment; is there no notion of "database" in Cassandra?

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/SourceInfo.java
```diff
@@ -28,49 +27,26 @@
...
-    public final String cluster;
-    public final OffsetPosition offsetPosition;
-    public final KeyspaceTable keyspaceTable;
-    public final boolean snapshot;
-    public final long tsMicro;
+    public String cluster;
+    public OffsetPosition offsetPosition;
+    public KeyspaceTable keyspaceTable;
+    public boolean snapshot;
+    public long tsMicro;
```
gunnarmorling: This should be `Instant` better. We try to convert any timestamps to `Instant` right at the system boundary as early as possible and then use that more meaningful type internally. Only where needed, it's "rendered" as long timestamp again.

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/SourceInfo.java
```diff
@@ -108,4 +84,19 @@ public String toString() {
         map.put(TIMESTAMP_KEY, tsMicro);
         return map.toString();
     }
+
+    @Override
+    protected Instant timestamp() {
+        return Instant.ofEpochSecond(tsMicro/1000);
```
gunnarmorling: See above, also should be converted keeping the precision.

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
bingqinzhou: > This shouldn't be needed, or do we actually want to maintain two different versions of the source struct? I don't think so, given the connector's "incubating" state. The notion of the "legacy" struct maker exists for the other connectors, as we did some technically non-backwards compatible changes to `source` in 0.10 (for the sake of general consistency and stream-lining, this might btw. relevant for you when upgrading your MySQL connectors) and we wanted to have a transitioning period where the old one still is available, too. This shouldn't be needed for Cassandra.

Understood, updated accordingly!

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/LegacyV1CassandraSourceInfoStructMaker.java
bingqinzhou: > See my comment above; shouldn't be needed.

Removed LegacyV1CassandraSourceInfoStructMaker accordingly.

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/SourceInfo.java
bingqinzhou: > Could you add a clarifying comment; is there no notion of "database" in Cassandra?

Got it, added clarifying comment accordingly.

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/SourceInfo.java
bingqinzhou: > See above, also should be converted keeping the precision.

Nice catch! Updated to use `Instant.ofEpochMilli(tsMicro)` to keep the precision.

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
bingqinzhou: > isEmitTombstoneOnDelete

We'll need to rewrite one here because the default value of `TOMBSTONES_ON_DELETE` in other Debezium Connectors is `true`, but in Cassandra Connector is `false`.

bingqinzhou:
> Hey @criccomini and @jgao54, any feedback from your side on this one?
> 
> @bingqinzhou, could you rebase it to the latest master, as there's a conflict. Thanks!

Sure, rebased and the conflict is resolved now :)

gunnarmorling:
> Rebased and the conflict is resolved now :)

Did you push this? Or maybe there's a new conflict (although I don't remember changing anything in Cassandra lately)?

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/SourceInfo.java
gunnarmorling: But `ofEpochMilli()` will loose precision, no?

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
gunnarmorling: Ah, would be nice to align the default then. Is this still something you wanted to do?

bingqinzhou: Make sense! Updated to align the default to be `true` as other debezium connectors. 
But want to double confirm with @jgao54 if it is ok to set the default value of `TOMBSTONES_ON_DELETE` to be `true` ..?

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/SourceInfo.java
bingqinzhou: Oops, my bad. Updated to convert `micro-seconds` to `Instant`.

bingqinzhou: > But `ofEpochMilli()` will loose precision, no?

Oops, my bad. Updated to convert `micro-seconds` to `Instant`.

bingqinzhou:
> > Rebased and the conflict is resolved now :)
> 
> Did you push this? Or maybe there's a new conflict (although I don't remember changing anything in Cassandra lately)?

@gunnarmorling Hm, are you still seeing any conflicts of this PR..? I resolved the conflicts 12 days ago (https://github.com/debezium/debezium-incubator/pull/109/commits/2ccb2cf14a329f4b9d521b0ff03d389b71c7f6d3) and pushed the change, currently I don't see there's any conflicts with this PR.

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
bingqinzhou: Make sense! Updated to align the default to be `true` as other debezium connectors. 
But want to double confirm with @jgao54 if it is ok to set the default value of `TOMBSTONES_ON_DELETE` to be `true` ..?

bingqinzhou: > Ah, would be nice to align the default then. Is this still something you wanted to do?

Make sense! Updated to align the default to be true as other debezium connectors.
But want to double confirm with @jgao54 if it is ok to set the default value of TOMBSTONES_ON_DELETE to be true ..?

jgao54: @bingqinzhou  @gunnarmorling 

Assume my understanding of Kafka tombstone and compaction is correct -- such that when a tombstone is detected, it will remove _all_ events in the log for the key _except for the most recent event with the same key_. If so, I'm a bit concerned with setting it to `true` as default, as it may result in unexpected data losses.

The reason is that the value of a CDC event may not contain all the columns, so the most recent event != the most recent row for a given key.

jgao54:
@bingqinzhou  sorry for the late response, will review this PR in more details by EOD today.

jgao54:
LGTM!

For data integrity, i think we should have default value for tombstone set to false, given users may not be checking every config when they initially launch the connector and if they have kafka log compaction turned on, they'd experience potential data loss.

bingqinzhou:
> LGTM!
> 
> For data integrity, i think we should have default value for tombstone set to false, given users may not be checking every config when they initially launch the connector and if they have kafka log compaction turned on, they'd experience potential data loss.

Make sense, reverted my previous commit and changed the default value of tombstone config to be false :)

bingqinzhou:
@gunnarmorling I kept the default value of `tombstoneOnDelete` to be `false` based on @jgao54's explanation about this. I've also made changes according to your previous comments as well as resolved the conflicts. If this looks good to you now, may I get this PR approved and merged please? Otherwise, please let me know if I need to make any further changes. Thank you!

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
gunnarmorling: Ah, yes. That's a good point. It seems `false` is the saver default then indeed.

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/SourceInfo.java
gunnarmorling: @bingqinzhou could you still address this remark?

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
gunnarmorling: In other words, the suggested change here is good AFAICS.

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/SourceInfo.java
bingqinzhou: @gunnarmorling Updated accordingly.

criccomini:
Checking in here. @gunnarmorling is this PR OK?

gunnarmorling:
Yes, it is from my PoV. Just needs a rebase. @bingqinzhou, once you've done that, feel free to merge it yourself or @criccomini can do it? In addition, an update of the connector docs in the core repo probably is needed to reflect the updated source structure (assuming there's an example of that shown in there).

gunnarmorling:
@bingqinzhou, @criccomini, so I went ahead and rebased this and squashed all into one commit. Could either of you double-check on this PR and merge it:

    https://github.com/debezium/debezium-incubator/pull/117

Thanks!

@bingqinzhou, btw. we're preferring rebases over merge commits for the sake of cleaner history. Let me know if you have any trouble with that in case you haven't used rebases before.

bingqinzhou:
> @bingqinzhou, @criccomini, so I went ahead and rebased this and squashed all into one commit. Could either of you double-check on this PR and merge it:
> 
> #117
> 
> Thanks!
> 
> @bingqinzhou, btw. we're preferring rebases over merge commits for the sake of cleaner history. Let me know if you have any trouble with that in case you haven't used rebases before.

@gunnarmorling I'm actually not sure how to rebase for a branch in a fork repo. Here's what I'm currently do to update with master:
```
$ git fetch upstream
$ git merge upstream/master
```
Please let me know if there's a better way that I'm supposed to use. Thank you!

PR #111, opened 2019-10-02
DBZ-1391 Do not allow code after closing brace

Naros:
https://issues.jboss.org/browse/DBZ-1391


gunnarmorling:
Thanks, @Naros!

jgao54:
Thanks!

gunnarmorling:
@Joy, to give some background: we're rolling out a more concise CheckStyle
sheet and also an auto-formatter which adheres to that.

It may sound scary at first, but actually we made very good experiences
with it in another project (Quarkus). It ends all those nitty-gritty
discussions around style which tend to add much stress without much value
(not that we had those too much in Debezium luckily, but still it's a
potential pain point). More time for everyone to focus on the more
interesting things than whitespace and indentation :)

PR #117, opened 2019-11-06
DBZ-1408 Refactor Cassandra SourceInfo to extend AbstractSourceInfo

gunnarmorling:
https://issues.jboss.org/browse/DBZ-1408

### DIFF
--- a/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
+++ b/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
@@ -12,16 +12,22 @@ import java.util.Optional;
 import java.util.Properties;
 import java.util.stream.Collectors;
 
+import io.debezium.config.Field;
+import io.debezium.config.Configuration;
+import io.debezium.config.CommonConnectorConfig;
+import io.debezium.connector.AbstractSourceInfo;
+import io.debezium.connector.SourceInfoStructMaker;
 import io.debezium.connector.cassandra.exceptions.CassandraConnectorConfigException;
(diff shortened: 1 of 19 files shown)


gunnarmorling:
@bingqinzhou, @criccomini, quick heads-up: something is wrong with the formatter config for the incubator repo, which is why this PR build failed. @Naros is looking into it and will merge it. The actual change is good, just some technicality due to the recently added auto-formatter. It will be part of the 1.0 Beta3 release.

gunnarmorling:
@bingqinzhou, while we're sorting out the formatter issue, could you please file another PR for the corresponding doc update? E.g. example events should reflect the new source structure. I also spotted one reference to "Avro schema", which shouldn't be there anymore as per the recent change to KC schema. Thanks!

jpechane:
@gunnarmorling Verified locally so it can be merged. The code needs to be remormatted

gunnarmorling:
Did the reformatting and applied. Many thanks, @bingqinzhou!

So on this formatting issue, we do apply automatic formatting of sources now, in order to avoid the fruitless debates about missing whitespace etc. This works very nice overall (happens as part of the build), but there's currently still one gotcha: the formatter config that's applied is located under _support/ide-configs/eclipse/debezium-formatter.xml_ in the main _debezium_ repo. The simplest to make it available to the incubator repo is a link like `ln -s ../debezium/support`. Or we copy the file into the incubator repo. If you're working with Eclipse, you canfigure the IDE also to apply the formatter when saving. Otherwise, if there are deviations from the formatting style, `mvn clean package` will apply the reformattings needed, leaving you with the changed files which you then can add to your commit. Let me know if you run into any issues with this.

PR #118, opened 2019-11-08
[DBZ-1594] Indicate if each column is a part of primary in kafka value

bingqinzhou:
https://issues.jboss.org/browse/DBZ-1594

Adding a pk field to indicate if a column is a part of primary key in Kafka value, as well as removing the set field from each column.

### DIFF
--- a/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CellData.java
+++ b/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CellData.java
@@ -44,7 +44,7 @@ public class CellData implements KafkaRecord {
 
     public static final String CELL_VALUE_KEY = "value";
     public static final String CELL_DELETION_TS_KEY = "deletion_ts";
-    public static final String CELL_SET_KEY = "set";
+    public static final String CELL_PK_KEY = "pk";
 
     public final String name;
     public final Object value;
@@ -67,7 +67,7 @@ public class CellData implements KafkaRecord {
         return new Struct(schema)
                 .put(CELL_VALUE_KEY, value)
                 .put(CELL_DELETION_TS_KEY, deletionTs)
-                .put(CELL_SET_KEY, true);
+                .put(CELL_PK_KEY, isPrimary());
     }
 
     static Schema cellSchema(ColumnMetadata cm, boolean optional) {
@@ -77,7 +77,7 @@ public class CellData implements KafkaRecord {
             SchemaBuilder schemaBuilder = SchemaBuilder.struct().name(cm.getName())
                     .field(CELL_VALUE_KEY, valueSchema)
                     .field(CELL_DELETION_TS_KEY, Schema.OPTIONAL_INT64_SCHEMA)
-                    .field(CELL_SET_KEY, Schema.BOOLEAN_SCHEMA);
+                    .field(CELL_PK_KEY, Schema.BOOLEAN_SCHEMA);
             if (optional) {
                 schemaBuilder.optional();
             }


gunnarmorling:
@bingqinzhou A few questions about this one:

* Why is the PK info needed within the message *values*? The PK columns already form the message *key*, so it seems that info would be redundant.
* On removing `set`, would be interesting to hear from @jgao54 what the intention for having this was. Just to make sure we don't miss anything.

bingqinzhou:
> @bingqinzhou A few questions about this one:
> 
> * Why is the PK info needed within the message _values_? The PK columns already form the message _key_, so it seems that info would be redundant.
> * On removing `set`, would be interesting to hear from @jgao54 what the intention for having this was. Just to make sure we don't miss anything.

Hi @gunnarmorling,

In order to do downstream de-duplication and compaction of data that's streamed into BigQuery from Cassandra change events, we need to know about the PK and non-PK columns of a Cassandra table. Currently the PK info is stored in Kafka key, which cannot be streamed into BigQuery through KCBQ. That's why we want to add this flag in Kafka value so that we can easily get PK info directly from BigQuery. 

For the `set` flag, I'm removing it because it's set to be `true` constantly at this moment. From my understanding, the intention of this `set` flag is to indicate if a column has been changed during a DML in Cassandra. But I feel this can alternatively be indicated by checking if the `deletion_ts` is not null (indicating that a column has been deleted) or the value is not null (indicating that a column has been updated or inserted). 

@jgao54 @criccomini Please correct me if I misunderstand the intention of `set` flag here, thank you!

jgao54:
@bingqinzhou  that's correct! the `set` flag is used to indicted changed columns, mainly so that consumer of these events can differentiate deleted columns from unmodified columns since both will have null values.

I don't think it is required though and your reasoning makes sense to me. I think the main benefit of  `set` column is making modified columns explicit, such that it's easier for consumer to filter out unnecessary data during view generation by ignore the unset ones, but is not required and can be derived as you suggested.

bingqinzhou:
@gunnarmorling Confirmed with @jgao54 that it's ok to remove `set` flag from kafka value schema, please refer to comments above.
@gunnarmorling @jgao54 May I get an approval for this PR if it looks good to you? Otherwise please let me if I need to make any additional changes :) We are waiting for this for downstream data compaction in BigQuery. Thank you!

jgao54:
LGTM. 

One other thought is that now that we are indicating primary key flag, and compaction option has been removed, it becomes safe to partition Cassandra CDC events by its partition key rather than primary key. The benefit would be that this makes Kafka mirror Cassandra's partition scheme, but im not sure if there are any practical benefit, just throwing the thought out there!

gunnarmorling:
Ah, that's an interesting idea, @jgao54. In that case I could get behind exposing the PK info in the message value. With the currently suggested form I'm a bit concerned about the duplication of information, it seems a bit like a work-around for the purposes of the BQ sink connector. Wouldn't it make more sense to adjust that one so it obtains the key info from the Kafka message key, @bingqinzhou?

bingqinzhou:
@gunnarmorling We're currently spiking to see if there's anyway we can change KCBQ to stream kafka keys into BQ. Will close this PR after we make changes from KCBQ.

gunnarmorling:
Hey @bingqinzhou, how is it going with your exploration of implementing this in KCBQ? Which option would make more sense for you?

bingqinzhou:
> Hey @bingqinzhou, how is it going with your exploration of implementing this in KCBQ? Which option would make more sense for you?

Hi @gunnarmorling We have a PR for refactoring KCBQ to consume and stream Kafka key now. I'm closing this PR here, thanks.

PR #123, opened 2019-11-11
DBZ-1555 TableId parts precedence is configurable

jpechane:
@gunnarmorling There is a core API compatibility break


Naros:
@jpechane thanks for spotting this, applied & thanks!

PR #125, opened 2019-11-21
DBZ-1631 - alter select to perform timestamp_to_scn on most recent

shelfoo:
Fix attempting to retrieve an SCN for all last_ddl_times, instead get most recent ddl_time and the SCN for that. Functionally equivalent, but won't fail when there's no snapshots for old ddls (on a pre-existing db for example)

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
@@ -158,7 +158,7 @@ public class OracleSnapshotChangeEventSource extends RelationalSnapshotChangeEve
             return Optional.empty();
         }
 
-        StringBuilder lastDdlScnQuery = new StringBuilder("SELECT MAX(TIMESTAMP_TO_SCN(last_ddl_time))")
+        StringBuilder lastDdlScnQuery = new StringBuilder("SELECT TIMESTAMP_TO_SCN(MAX(last_ddl_time))")
                 .append(" FROM all_objects")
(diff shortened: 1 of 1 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
```diff
@@ -158,7 +158,7 @@ private boolean areSameTimestamp(Long scn1, long scn2) throws SQLException {
             return Optional.empty();
         }
 
-        StringBuilder lastDdlScnQuery = new StringBuilder("SELECT MAX(TIMESTAMP_TO_SCN(last_ddl_time))")
+        StringBuilder lastDdlScnQuery = new StringBuilder("SELECT TIMESTAMP_TO_SCN(MAX(last_ddl_time))")
```
gunnarmorling: @jpechane, any thoughts?

jpechane: @shelfoo Looks fine for me, is there a way how to test the change?

shelfoo: @jpechane I can maybe get a green->green test in place, tough to do red->green without having a db that is old. Other than that just running this sql on any db proves that it has the same effect

jpechane: @shelfoo ok, let's do it without test. Thanks a lot, merging.

PR #129, opened 2020-01-28
DBZ-1052 Transaction metadata for Oracle

jpechane:
https://issues.redhat.com/browse/DBZ-1052

Depends on https://github.com/debezium/debezium/pull/1221


gunnarmorling:
Applied, thanks, @jpechane!

PR #131, opened 2020-01-31
[DBZ-1763] Replace connectorName with kafkaTopicPrefix for kafka key/value schema namespace.

bingqinzhou:
Use `kafkaTopicPrefix` which is actually the logical cluster name in kafka key/value namespace instead of `connectorName`, otherwise, different cassandra-cdc connectors deployed on different cassandra nodes will produce different kafka key/value schemas for one same kafka topic and cause downstream kcbq to fail.


gunnarmorling:
Thanks, @bingqinzhou! Just was thinking of you folks and how it may go with the connector for you folks :)

Could you file a JIRA issue and change the commit message to refer to its number? And can you update the examples in the connector documentation (in the main repo) to reflect that change? Also wondering whether there should be a test related to that? Thanks!

bingqinzhou:
> Thanks, @bingqinzhou! Just was thinking of you folks and how it may go with the connector for you folks :)
> 
> Could you file a JIRA issue and change the commit message to refer to its number? And can you update the examples in the connector documentation (in the main repo) to reflect that change? Also wondering whether there should be a test related to that? Thanks!

Looks like I messed up the PR when trying to edit previous commit messages. I'm closing this PR and created https://github.com/debezium/debezium-incubator/pull/132 to address the changes :)

PR #132, opened 2020-02-04
[DBZ-1763] Replace connectorName with kafkaTopicPrefix for kafka key/value schema namespace.

bingqinzhou:
Use `kafkaTopicPrefix` which is actually the logical cluster name in kafka key/value namespace instead of `connectorName`, otherwise, different cassandra-cdc connectors deployed on different cassandra nodes will produce different kafka key/value schemas for one same kafka topic and cause downstream kcbq to fail.

### DIFF
--- a/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CassandraConnectorContext.java
+++ b/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CassandraConnectorContext.java
@@ -35,7 +35,7 @@ public class CassandraConnectorContext {
         this.queue = new BlockingEventQueue<>(this.config.pollIntervalMs(), this.config.maxQueueSize(), this.config.maxBatchSize());
 
         // Setting up schema holder ...
-        this.schemaHolder = new SchemaHolder(this.cassandraClient, this.config.connectorName(), this.config.getSourceInfoStructMaker());
+        this.schemaHolder = new SchemaHolder(this.cassandraClient, this.config.kafkaTopicPrefix(), this.config.getSourceInfoStructMaker());
 
         // Setting up a file-based offset manager ...
         this.offsetWriter = new FileOffsetWriter(this.config.offsetBackingStoreDir());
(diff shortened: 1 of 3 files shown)


bingqinzhou:
Screen Shot 2020-02-03 at 5 19 52 PM
Build failure doesn't seem to be related to changes in this PR..?

gunnarmorling:
That's interesting, definitely not related to your change. @jpechane, was there any CI job that was failing due to that?

gunnarmorling:
@bingqinzhou, can you refresh my mind -- is the Cassandra connector still using Avro directly, or is it creating Kafka Connect ``Struct``s by now? I thought it'd be the latter, but the docs still speak about Avro, also the notion of "namespace" shouldn't be present when using JSON serialization of Connect records. It seems I'm missing something atm. Thanks for clarifying!

bingqinzhou:
> @bingqinzhou, can you refresh my mind -- is the Cassandra connector still using Avro directly, or is it creating Kafka Connect `Struct`s by now? I thought it'd be the latter, but the docs still speak about Avro, also the notion of "namespace" shouldn't be present when using JSON serialization of Connect records. It seems I'm missing something atm. Thanks for clarifying!

@gunnarmorling Yes Cassandra connector has been using Kafka Connect Schema instead of Avro Schema for a while, will update the inappropriate "Avro" parts in the documentation. I wasn't aware of that "namespace" shouldn't be present for JSON serialized records, what would be a good way to handle this please? Since this sounds like a different issue from the intent of DBZ-1763, I created DBZ-1765 to track it. Thank you!

gunnarmorling:
@bingqinzhou, see my comment on DBZ-1765, ideally the the structure created by the Cassandra connector looks like the one created by the others. We can handle it separately, as you say.

Can you rebase this PR's branch and force push it, this will trigger another build (the failure you saw has been addressed in between). Thanks!

bingqinzhou:
> @bingqinzhou, see my comment on DBZ-1765, ideally the the structure created by the Cassandra connector looks like the one created by the others. We can handle it separately, as you say.
> 
> Can you rebase this PR's branch and force push it, this will trigger another build (the failure you saw has been addressed in between). Thanks!

Rebased accordingly!

jpechane:
Superseded with https://github.com/debezium/debezium-incubator/pull/134

PR #134, opened 2020-02-05
DBZ-1763 Replace connectorName with kafkaTopicPrefix

jpechane:


jpechane:
Applied, thanks @bingqinzhou!

PR #135, opened 2020-02-07
DBZ-1752 DB2 connector follow-up tasks

lga-zurich:
This pull request addresses:
- Use only DB2 compatible isolation modes when taking Snapshots
- Remove some unused variables

### DIFF
--- a/debezium-connector-db2/src/main/java/io/debezium/connector/db2/Db2ConnectorConfig.java
+++ b/debezium-connector-db2/src/main/java/io/debezium/connector/db2/Db2ConnectorConfig.java
@@ -37,7 +37,7 @@ import io.debezium.relational.history.KafkaDatabaseHistory;
 /**
  * The list of configuration options for DB2 connector
  *
- * @author Jiri Pechanec
+ * @author Jiri Pechanec, Luis Garcés-Erice
  */
 public class Db2ConnectorConfig extends HistorizedRelationalDatabaseConnectorConfig {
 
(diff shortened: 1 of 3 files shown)


gunnarmorling:
LGTM, thanks! Can you send a PR for updating the docs in regards to the locking option, too?

In terms of releases: we did 1.1 Beta1 yesterday (announcement still pending), so this will be in the next 1.1 preview release (Beta2 or CR2, still to be decided). These releases happen every three weeks give or take, and there should be a new minor (1.1, 1.2, etc.) every quarter.

gunnarmorling:
Ah, one other thing: could you please always prefix commit messages with the issue key (DBZ-1752...) and avoid merge commits (i.e. prefer rebases, for the sake of cleaner history). It's fine to force-push to a PR branch for updating such things (in this case here I'd squash everything into one anyways, as e.g. the formatting commit doesn't add much value in the history in the long term). Thanks!

lga-zurich:
> Ah, one other thing: could you please always prefix commit messages with the issue key (DBZ-1752...) and avoid merge commits (i.e. prefer rebases, for the sake of cleaner history). It's fine to force-push to a PR branch for updating such things (in this case here I'd squash everything into one anyways, as e.g. the formatting commit doesn't add much value in the history in the long term). Thanks!

Yes, I always forget, done!

lga-zurich:
> LGTM, thanks! Can you send a PR for updating the docs in regards to the locking option, too?

There you go debezium/debezium#1246

gunnarmorling:
Applied; thanks a lot, @lga-zurich!

PR #143, opened 2020-02-27
[DBZ-1820] Replace BlockingEventQueue with Debezium ChangeEventQueue.

bingqinzhou:
Use built-in ChangeEventQueue in Debezium instead of self-implemented BlockingEventQueue.


debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CommitLogProcessor.java
```diff
@@ -102,15 +105,21 @@ void processCommitLog(File file) throws IOException {
             throw new IOException("Commit log " + file.getName() + " does not exist");
         }
         try {
-            LOGGER.info("Processing commit log {}", file.getName());
-            metrics.setCommitLogFilename(file.getName());
-            commitLogReader.readCommitLogSegment(commitLogReadHandler, file, false);
-            queue.enqueue(new EOFEvent(file, true));
-            LOGGER.info("Successfully processed commit log {}", file.getName());
+            try {
```
gunnarmorling: Is this "inner" try/catch block really needed? Couldn't botch catch statements just be part of the same try/catch block?

debezium-connector-cassandra/src/test/java/io/debezium/connector/cassandra/SnapshotProcessorTest.java
```diff
@@ -89,17 +91,17 @@ public void testSnapshotEmptyTable() throws Exception {
         context.getCassandraClient().execute("CREATE TABLE IF NOT EXISTS " + keyspaceTable("cdc_table") + " (a int, b text, PRIMARY KEY(a)) WITH cdc = true;");
         context.getSchemaHolder().refreshSchemas();
 
-        BlockingEventQueue<Event> queue = context.getQueue();
-        assertTrue(queue.isEmpty());
+        ChangeEventQueue<Event> queue = context.getQueue();
+        assertTrue(queue.remainingCapacity() == queue.totalCapacity());
```
gunnarmorling: Better use `assertEquals()`; in case of a failure the message will reveal expected and actual values.

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/RecordMaker.java
```diff
@@ -16,26 +19,27 @@
     private final boolean emitTombstoneOnDelete;
     private final Filters filters;
     private final SourceInfo sourceInfo;
+    private static final Logger LOGGER = LoggerFactory.getLogger(RecordMaker.class);
```
gunnarmorling: Could you put the `static` field before all the instance fields?

debezium-connector-cassandra/src/test/java/io/debezium/connector/cassandra/SnapshotProcessorTest.java
bingqinzhou: Updated accordingly~

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/RecordMaker.java
bingqinzhou: Updated accordingly~

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CommitLogProcessor.java
bingqinzhou: > Is this "inner" try/catch block really needed? Couldn't botch catch statements just be part of the same try/catch block?

Yes I think so, because the inner catch block which catches the `IOException` will also throw an `InterruptedException`.

```catch (IOException e) {
                queue.enqueue(new EOFEvent(file, false));
                LOGGER.warn("Error occurred while processing commit log " + file.getName(), e);
            }```

gunnarmorling: Gotcha, makes sense.

PR #144, opened 2020-03-04
DBZ-1842 Include Db2 in root incubator README

lga-zurich:


gunnarmorling:
Thanks, @lga-zurich. I don't think we should duplicate the entire contents of the connector-specific README in the root one, though.

Taking a step back, this is part of the more general discussion happening in [DBZ-1754](https://issues.redhat.com/browse/DBZ-1754). Perhaps that's something you'd be interested to tackle in a more holistic way? If not, I'd suggest we just add the minimum of info to the incubator root README on how to build the Db2 connector for now, and linking to the one in the sub-dir for more details.

Eventually, as described in DBZ-1754, I think we should abandon all these detailed READMEs and come up with a development guide containing that info in a well-organized and redundancy-free way.

lga-zurich:
@gunnarmorling ok, got you. I have substantially reduced the Db2 section in the root README.md to one paragraph. I just wanted something there telling you that "yes, you downloaded the Db2 connector", because when I did it and only found Oracle documentation I had to actually check which class files where included in the jar to make sure they were the right ones 😄 

Now it just links to the README in the connector directory. As for DBZ-1754, that's probably the right thing to do, but for now I just want to minimize the confusion.

gunnarmorling:
Thx! I think it's much better now, until DBZ-1754 has been addressed more generally.

PR #149, opened 2020-03-20
DBZ-1892 Fix wrong sourceInfo in Cassandra CDC Kafka messages

bingqinzhou:
The PR is for fixing a critical bug that some change event or snapshot kafka messages published from Cassandra Connector for one table include wrong source info from another table.


gunnarmorling:
LGTM, @bingqinzhou. Perhaps you could add a test which asserts (parts of) the resulting source info in change events?

jpechane:
Added test in https://github.com/debezium/debezium-incubator/pull/150

PR #153, opened 2020-04-01
[DBZ-1407] Use Debezium TopicSelector

bingqinzhou:
Replace custom CassandraTopicSelector with Debezium TopicSelector.


gunnarmorling:
Squashed and merged. Thanks, @bingqinzhou!

PR #154, opened 2020-04-01
[DBZ-1927] Do not remove processed commitLog files under testing mode.

bingqinzhou:
Cassandra Connector shouldn't remove commitLog files under testing mode (`latest.commit.log.only=true`).

Added a condition check to only enqueue EOF (end-of-file) event when not under testing mode, so that QueueProcess won't poll and process a corresponding EOF event to move the commitLog file to relocation directory.


gunnarmorling:
Thanks, @bingqinzhou! Btw. are you using the open-sourced version of this connector already in prod? // CC @criccomini

bingqinzhou:
> Thanks, @bingqinzhou! Btw. are you using the open-sourced version of this connector already in prod? // CC @criccomini

@gunnarmorling No, we have only deployed Cassandra Connector on POC currently.

PR #156, opened 2020-04-14
[DBZ-1967] Fix the error for deserializing column mutation with reversed type.

bingqinzhou:
https://issues.redhat.com/browse/DBZ-1967

For a column with DDL like `WITH CLUSTERING ORDER BY (event_time DESC)`, the abstract type of this column is wrapped with a ReversedType outside like
`org.apache.cassandra.db.marshal.ReversedType(org.apache.cassandra.db.marshal.TimestampType)`, which causes deserialization to fail thus breaks commitLog processing in Cassandra Connector.

- Check if a type passed in for deserialization is a reversed type, if yes, use it's base type for deserialization.
- Add more detailed logs for debugging which column and type is failed to be deserialized.

### DIFF
--- a/debezium-connector-cassandra/src/test/java/io/debezium/connector/cassandra/transforms/CassandraTypeDeserializerTest.java
+++ b/debezium-connector-cassandra/src/test/java/io/debezium/connector/cassandra/transforms/CassandraTypeDeserializerTest.java
@@ -40,6 +40,7 @@ import org.apache.cassandra.db.marshal.Int32Type;
 import org.apache.cassandra.db.marshal.ListType;
 import org.apache.cassandra.db.marshal.LongType;
 import org.apache.cassandra.db.marshal.MapType;
+import org.apache.cassandra.db.marshal.ReversedType;
 import org.apache.cassandra.db.marshal.SetType;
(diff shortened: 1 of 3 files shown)


debezium-connector-cassandra/src/test/java/io/debezium/connector/cassandra/transforms/CassandraTypeDeserializerTest.java
```diff
@@ -425,4 +426,18 @@ public void testUUIDType() {
 
         Assert.assertEquals(expectedFixedUUID, deserializedUUID);
     }
+
+    @Test
+    public void testReversedType() {
```
gunnarmorling, line 431: Could you add the `@FixFor` annotation?

gunnarmorling:
Added the `@FixFor` and applied (got a release coming up tonight, so I thought I'd quickly wrap it up). Thanks, @bingqinzhou!

bingqinzhou:
> Added the `@FixFor` and applied (got a release coming up tonight, so I thought I'd quickly wrap it up). Thanks, @bingqinzhou!

Thank you @gunnarmorling!

PR #160, opened 2020-04-23
DBZ-1878 Use Oracle driver from central

jpechane:
https://issues.redhat.com/browse/DBZ-1878

### DIFF
--- a/debezium-connector-oracle/pom.xml
+++ b/debezium-connector-oracle/pom.xml
@@ -26,7 +26,7 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
-            <groupId>com.oracle.instantclient</groupId>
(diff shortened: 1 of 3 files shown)


debezium-connector-oracle/pom.xml
```diff
@@ -26,7 +26,7 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
-            <groupId>com.oracle.instantclient</groupId>
+            <groupId>com.oracle.database.jdbc</groupId>
```
gunnarmorling, line 29: To be sure: it won't end up in the dist TAR.GZ, right?

jpechane: Tested, it is not there.

PR #162, opened 2020-04-26
DBZ-1591 Add parallel snapshot

grantcooksey:
https://issues.redhat.com/browse/DBZ-1591

Adds support for performing snapshots concurrently to the postgres and oracle
connectors. The snapshot process is now performed using a fixed worker thread
pool whose size is controlled by the "snapshot.max.threads" property.
Worker threads open a new connection to the datasource to snapshot a single
datasource object(ex. table) and close when finished. Connectors that do
support concurrent snapshots is a special case where the existing connection
is reused, performed on a fixed single thread pool, and all snapshotted objects
are published in order.

Also see: https://github.com/debezium/debezium/pull/1450


gunnarmorling:
Closing this one for now, as it's outdated and we're going to revisit snapshot parallelizing more generally soon. Thanks for your efforts nevertheless, @grantcooksey! Will ping you as soon as this is getting more specific.

PR #163, opened 2020-04-29
DBZ-1985 More meaningful exception in case of non-existent commit log…

gunnarmorling:
… dir

https://issues.redhat.com/browse/DBZ-1985

Hey @bingqinzhou, so I had another look at the NPE. The only way that `getCommitLogs()` returns null (which in turn causes the NPE in `sort()`) is if the given directory doesn't exist. So I'm raising a more explicit exception in this case. Could you review and merge? // CC @criccomini @jgao54

### DIFF
--- a/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CommitLogUtil.java
+++ b/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CommitLogUtil.java
@@ -69,6 +69,10 @@ public final class CommitLogUtil {
      * If the directory does not contain any commit logs, an empty array is returned.
      */
     public static File[] getCommitLogs(File directory) {
+        if (!directory.isDirectory()) {
+            throw new IllegalArgumentException("Given directory does not exist: " + directory);
+        }
+
         return directory.listFiles(f -> f.isFile() && FILENAME_REGEX_PATTERN.matcher(f.getName()).matches());
     }
 


debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CommitLogUtil.java
```diff
@@ -69,6 +69,10 @@ public static void deleteCommitLog(File file) {
      * If the directory does not contain any commit logs, an empty array is returned.
      */
     public static File[] getCommitLogs(File directory) {
+        if (!directory.isDirectory()) {
```
jgao54, line 72: oof, good catch!

Looking at [java doc on listFiles](https://docs.oracle.com/javase/6/docs/api/java/io/File.html#listFiles%28%29):
```
Returns:
An array of abstract pathnames denoting the files and directories in the directory denoted by this abstract pathname. The array will be empty if the directory is empty. Returns null if this abstract pathname does not denote a directory, or if an I/O error occurs.
```

jgao54:
It sounds like this NPE issue was a transient error and could not reproduced again. If so, I think the root root cause may be due to Cassandra configs were not initialized properly [here] and the configs did not get loaded properly (https://github.com/debezium/debezium-incubator/blob/v1.1.0.Final/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CassandraConnectorContext.java#L63-L65). 

I can see this fails if the Daemon or Client was already initialized. This is my bad, i think the error handling here is lacking.

jgao54:
Nvm, scratch that, by default the argument passed in to toolInitialization is `true`, which means it _should_ throw an exception if either client or daemon was initialized :(

jgao54:
I re-read the original issue thread, looks like the issue is in the Post Processor,  which requires `COMMIT_LOG_RELOCATION_DIR` to be set (this PR is still an improvement, as it will surface the missing directory since it uses the same `getCommitLogs` methods )

To provide some background context,  once the Cassandra connector processes a commit log, it move the file to the relocation directory for post-processing. And by default the post-processor will just deletes the file (this is to prevent eating up local disk space as commit logs accumulates), but can be subclassed to do things like uploading to remote storage if needed.

Right now post-processing is enabled by default but we don't throw an exception if the relocation directory is not supplied. My hunch is this is what caused the NPE.

gunnarmorling:
Ok, sweet. Thanks for merging and providing some more context, @jgao54!

gunnarmorling:
Will resolve DBZ-1985 with that. Thanks, all!

bingqinzhou:
@gunnarmorling Sorry for my late response, thank you and @jgao54 for the insights and improvement! Understood how the NPE happened now :) 

Since `CommitLogPostProcessor` has a dependency on the `"commit.log.relocation.dir"` config, would we want to check if `"commit.log.relocation.dir"` is set and valid when initializing `CommitLogPostProcessor`, and if not, raise an exception?

jgao54:
Thanks @bingqinzhou! I think it would be good to check. So both `CommitLogProcessor` and `CommitLogPostProcessor` use it via the shared connector context, it may be better to do the check upon start up, maybe to have a logic like "if `commit.log.post.processing.enabled` is set to true, check to make sure relocation dir is set and valid"

PR #165, opened 2020-05-05
DBZ-2051 Replace initial_schema_only with schema_only

Naros:
https://issues.redhat.com/browse/DBZ-2051

### DIFF
--- a/debezium-connector-db2/src/main/java/io/debezium/connector/db2/Db2ConnectorConfig.java
+++ b/debezium-connector-db2/src/main/java/io/debezium/connector/db2/Db2ConnectorConfig.java
@@ -54,7 +54,7 @@ public class Db2ConnectorConfig extends HistorizedRelationalDatabaseConnectorCon
         /**
          * Perform a snapshot of the schema but no data upon initial startup of a connector.
          */
-        INITIAL_SCHEMA_ONLY("initial_schema_only", false);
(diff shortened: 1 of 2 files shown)


debezium-connector-db2/src/main/java/io/debezium/connector/db2/Db2ConnectorConfig.java
```diff
@@ -51,10 +51,18 @@
          */
         INITIAL("initial", true),
 
+        /**
+         * Perform a snapshot of the schema but no data upon initial startup of a connector.
+         *
+         * @deprecated to be removed in 1.3; use {@link #SCHEMA_ONLY} instead.
+         */
+        @Deprecated
```
Naros, line 59: The documentation for `snapshot.mode` did not include any reference to `initial_schema_only` so I'm wondering if we should just go ahead and remove this rather than deprecating and logging a warning.  I think both are fine and the latter would be the least disruptive for users who might be using this connector already.  Whatever you think is best @gunnarmorling.

gunnarmorling: Ah, good catch. I think at this point we simply can change this, without deprecating first. It's one of the reasons for the connector being in "incubation" state, giving us more freedom for rectifying things like this.

gunnarmorling:
Applied, thx.

PR #167, opened 2020-05-15
DBZ-2066 Add more logs

crazy-2020:
Add logs when the table snapshot is taken and when emitting a kafka message for a single record/mutation


debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/SnapshotProcessor.java
```diff
@@ -155,6 +155,7 @@ private void takeTableSnapshot(TableMetadata tableMetadata) throws IOException {
         LOGGER.info("Executing snapshot query '{}' with consistency level {}", statement.getQueryString(), statement.getConsistencyLevel());
         ResultSet resultSet = cassandraClient.execute(statement);
         processResultSet(tableMetadata, resultSet);
+        LOGGER.debug("The snapshot of table '{}' has been taken", tableName(tableMetadata));
```
gunnarmorling, line 158: Related to that, it'd also be nice to look into metrics. We'd get those "for free", when moving the Cassandra connector to the framework used by the other Debezium connectors :)

gunnarmorling:
Applied, thanks! @crazy-2020, what name should I add for you to the list of all authors?

PR #170, opened 2020-05-19
[DBZ-2091] Fix errors encountered for User-Defined-Type data

bingqinzhou:
The PR fixes the following issues:

- Fails to convert UDT type schema.
- Casandra Connection is not closed when connector context fails to be initialized.

### DIFF
--- a/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CellData.java
+++ b/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CellData.java
@@ -83,6 +83,9 @@ public class CellData implements KafkaRecord {
         return struct;
     }
 
+    // Encountered DataException("Struct schemas do not match.") when cell value is a Struct,
+    // because equal() for Kafka Connect Schema is using the default shallow comparison.
+    // Replace the schema of value struct with the same schema object passed in to avoid this problem.
     private Struct cloneValue(Schema valueSchema, Object value) {
         Struct valueStruct = (Struct) value;
         Struct clonedValue = new Struct(valueSchema);
--- a/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CellData.java
+++ b/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CellData.java
@@ -66,10 +66,16 @@ public class CellData implements KafkaRecord {
 
     @Override
     public Struct record(Schema schema) {
-        return new Struct(schema)
-                .put(CELL_VALUE_KEY, value)
-                .put(CELL_DELETION_TS_KEY, deletionTs)
-                .put(CELL_SET_KEY, true);
+        Struct cellDataStruct;
+        if(value instanceof Struct) {
+            cellDataStruct = new Struct(((Struct) value).schema());
+        } else {
+            cellDataStruct = new Struct(schema);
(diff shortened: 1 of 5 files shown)


gunnarmorling:
Hey @bingqinzhou, the test failure on CI looks related. Also, could you add a test for the specific issue being fixed here? Thanks!

bingqinzhou:
> Hey @bingqinzhou, the test failure on CI looks related. Also, could you add a test for the specific issue being fixed here? Thanks!

Yea, working on fixing the test currently

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CellData.java
```diff
@@ -66,10 +67,33 @@ public boolean isPrimary() {
...
+        }
+
+        return cellStruct;
+    }
+
+    // Encountered DataException("Struct schemas do not match.") when cell value is a Struct,
+    // because equal() for Kafka Connect Schema is using the default shallow comparison.
+    // Replace the schema of value struct with the same schema object passed in to avoid this problem.
+    private Struct cloneValue(Schema valueSchema, Object value) {
+        Struct valueStruct = (Struct) value;
```
gunnarmorling, line 90: Better declare `value` a `Struct` already, pushing the cast to the method call site, where it's known to be a struct.

gunnarmorling: Could we also encounter a struct of struct, i.e. does this method have to be called recursively?

bingqinzhou:
> LGTM, AFAICS. Two comments inline; also I still think it'd be nice to have a test show-casing the actual issue that's fixed here (i.e. one which would have failed before).

The test has already covered the issue we're trying to fix. It passed and failed to catch this issue before because the UserType is mocked in this test and we make `userType.getTypeArguments()` return a list of abstract type, which however returns an empty list in real.

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CellData.java
bingqinzhou: > Better declare `value` a `Struct` already, pushing the cast to the method call site, where it's known to be a struct.

Updated accordingly :)

bingqinzhou: > Could we also encounter a struct of struct, i.e. does this method have to be called recursively?

Nice catch! I tried to process a nested UDT data in cassandra and I was wrong about the cause of `DataException("Struct schemas do not match.")` here. Kafka Connect Schema does have an implemented equal() to compare schemas, the error is instead because the `valueSchema` is `optional`, but the schema of `value` formed during deserialization is not. Since the ultimate fix for this might touch a lot of places, I keep cloneValue() here for a temporary fix and created https://issues.redhat.com/browse/DBZ-2103 separately for this issue.

gunnarmorling: Gotcha, makes sense!

gunnarmorling:
> It passed and failed to catch this issue before because the UserType is mocked in this test...

Hehe, I see. No reason of mocks for pretty much that reason ;)

PR #173, opened 2020-06-05
[DBZ-2117] Fix ttl() and write_time() error for collection-type column during snapshot

bingqinzhou:
https://issues.redhat.com/browse/DBZ-2117

Cassandra Connector cannot capture delete events when TTL passes for collection-type column at this moment because of the restriction in Cassandra. 

To avoid the errors encountered during snapshot for collection-type columns:
- Do not call ttl() on collection-type columns during snapshot.
- Do not call write_time() on any columns during snapshot, use `execution_time` instead of `write_time` for `source.ts_ms` for snapshot events.

### DIFF
--- a/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/SnapshotProcessor.java
+++ b/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/SnapshotProcessor.java
@@ -7,6 +7,7 @@ package io.debezium.connector.cassandra;
 
 import java.io.IOException;
 import java.time.Duration;
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
@@ -50,6 +51,7 @@ public class SnapshotProcessor extends AbstractProcessor {
     private static final String NAME = "Snapshot Processor";
     private static final String CASSANDRA_NOW_UNIXTIMESTAMP = "UNIXTIMESTAMPOF(NOW())";
     private static final String EXECUTION_TIME_ALIAS = "execution_time";
+    private static final Set<DataType.Name> collectionTypes = new HashSet<>(Arrays.asList(DataType.Name.LIST, DataType.Name.SET, DataType.Name.MAP));
 
     private final CassandraClient cassandraClient;
     private final ChangeEventQueue<Event> queue;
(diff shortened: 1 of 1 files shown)


debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/SnapshotProcessor.java
```diff
@@ -50,6 +51,7 @@
     private static final String NAME = "Snapshot Processor";
     private static final String CASSANDRA_NOW_UNIXTIMESTAMP = "UNIXTIMESTAMPOF(NOW())";
     private static final String EXECUTION_TIME_ALIAS = "execution_time";
+    private static final Set<DataType.Name> collectionTypes = new HashSet<>(Arrays.asList(DataType.Name.LIST, DataType.Name.SET, DataType.Name.MAP));
```
gunnarmorling, line 54: Using `io.debezium.util.Collect` will simplify this a bit.

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/SnapshotProcessor.java
```diff
@@ -209,12 +212,12 @@ private void processResultSet(TableMetadata tableMetadata, ResultSet resultSet)
         while (rowIter.hasNext()) {
             if (isRunning()) {
                 Row row = rowIter.next();
-                WriteTimeHolder writeTimeHolder = new WriteTimeHolder();
-                RowData after = extractRowData(row, tableMetadata.getColumns(), partitionKeyNames, clusteringKeyNames, writeTimeHolder);
+                Object executionTime = readExecutionTime(row);
```
gunnarmorling, line 215: Seems this could be `long` instead of `Object` as per the cast further below?

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/SnapshotProcessor.java
```diff
@@ -234,27 +237,20 @@ private void processResultSet(TableMetadata tableMetadata, ResultSet resultSet)
     /**
      * This function extracts the relevant row data from {@link Row} and updates the maximum writetime for each row.
      */
-    private static RowData extractRowData(Row row, List<ColumnMetadata> columns, Set<String> partitionKeyNames, Set<String> clusteringKeyNames,
-                                          WriteTimeHolder writeTimeHolder) {
+    private static RowData extractRowData(Row row, List<ColumnMetadata> columns, Set<String> partitionKeyNames, Set<String> clusteringKeyNames, Object executionTime) {
```
gunnarmorling, line 240: `long`?

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/SnapshotProcessor.java
bingqinzhou: Hm, it seems that executionTime can possibly be null, thus we don't want to cast it into long before verifying that it's not null..?
https://github.com/debezium/debezium-incubator/blob/master/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/SnapshotProcessor.java#L250

bingqinzhou: > Seems this could be `long` instead of `Object` as per the cast further below?

Hm, it seems that executionTime can possibly be null, thus we don't want to cast it into long before verifying that it's not null..?
https://github.com/debezium/debezium-incubator/blob/master/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/SnapshotProcessor.java#L250

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/SnapshotProcessor.java
bingqinzhou: Updated accordingly :)

bingqinzhou:
Hm, build failure doesn't seem to be related to changes in this PR:

bingqinzhou:
Hm, build failure doesn't seem to related to changes in this PR:
Screen Shot 2020-06-12 at 11 25 44 AM

gunnarmorling:
Yes, that's indeed unrelated. Could you please rebase to latest master and force-push?

gunnarmorling:
Squashed and applied. Thanks, @bingqinzhou!

PR #180, opened 2020-07-14
[DBZ-2103] Ultimate fix for data mismatch error.

bingqinzhou:
Final fix for issue described in https://issues.redhat.com/browse/DBZ-2103 by making all column type optional consistently.

### DIFF
--- a/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CellData.java
+++ b/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CellData.java
@@ -8,7 +8,6 @@ package io.debezium.connector.cassandra;
 import java.util.Objects;
 
 import org.apache.cassandra.db.marshal.AbstractType;
-import org.apache.kafka.connect.data.Field;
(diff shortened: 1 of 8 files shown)


bingqinzhou:
@gunnarmorling May I get a review for PR? Please let me know if you have any concern about this change. Thank you!

gunnarmorling:
@bingqinzhou, so are those values always optional actually? I.e. there's no way to declare them as _NOT NULL_ as in SQL?

bingqinzhou:
> @bingqinzhou, so are those values always optional actually? I.e. there's no way to declare them as _NOT NULL_ as in SQL?

@gunnarmorling I don't think we should force any value schema to be non-opitional:
- There doesn't seem to be a `NOT NULL` constraint for Cassandra columns.
- Even if the value of a column is not supposed to be `null`, if the column is `deleted` or not included in an `update` operation of the row, the deserialized value of this column (from it's corresponding mutation in commitLog) should still be `null` in kafka message.

gunnarmorling:
Ok, that makes sense. Thanks! Will merge.

PR #183, opened 2020-08-14
[DBZ-2410] Improve error handling for Cassandra Connector

bingqinzhou:
- Catch all exceptions when processing commitLogFiles instead of just IOException, if users use their self-implemented CommitLogTransfer class instead of the default BlackHoleCommitLogTransfer, Cassandra Connector doesn't need to stop for commitLog processing errors.
- Enable Cassandra Connector to re-process error commitLogFiles when related config is enabled.

### DIFF
--- a/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CommitLogProcessor.java
+++ b/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CommitLogProcessor.java
@@ -125,10 +125,11 @@ public class CommitLogProcessor extends AbstractProcessor {
                 if (!latestOnly) {
                     queue.enqueue(new EOFEvent(file, false));
                 }
-                LOGGER.error("Error occurred while processing commit log " + file.getName(), e);
                 if (commitLogTransfer.getClass().getName().equals(CassandraConnectorConfig.DEFAULT_COMMIT_LOG_TRANSFER_CLASS)) {
+                    LOGGER.error("Error occurred while processing commit log " + file.getName(), e);
                     throw e;
                 }
+                LOGGER.warn("Error occurred while processing commit log " + file.getName(), e);
             }
         }
         catch (InterruptedException e) {
--- a/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CommitLogProcessor.java
+++ b/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CommitLogProcessor.java
@@ -40,6 +40,8 @@ public class CommitLogProcessor extends AbstractProcessor {
     private final boolean latestOnly;
     private final CommitLogProcessorMetrics metrics = new CommitLogProcessorMetrics();
     private boolean initial = true;
+    private final boolean errorCommitLogReprocessEnabled;
+    private final CommitLogTransfer commitLogTransfer;
 
     public CommitLogProcessor(CassandraConnectorContext context) throws IOException {
         super(NAME, 0);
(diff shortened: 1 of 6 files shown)


debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CommitLogProcessor.java
```diff
@@ -115,11 +121,14 @@ void processCommitLog(File file) throws IOException {
                 }
                 LOGGER.info("Successfully processed commit log {}", file.getName());
             }
-            catch (IOException e) {
+            catch (Exception e) {
                 if (!latestOnly) {
                     queue.enqueue(new EOFEvent(file, false));
                 }
                 LOGGER.warn("Error occurred while processing commit log " + file.getName(), e);
+                if (commitLogTransfer.getClass().toString() == CassandraConnectorConfig.DEFAULT_COMMIT_LOG_TRANSFER_CLASS) {
```
jpechane, line 129: Do we really want to use equality operator here? Also `DEFAULT_COMMIT_LOG_TRANSFER_CLASS` contians the name of the calss but the `toString` here is called the `class` itself.

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CommitLogProcessor.java
```diff
@@ -115,11 +121,14 @@ void processCommitLog(File file) throws IOException {
                 }
                 LOGGER.info("Successfully processed commit log {}", file.getName());
             }
-            catch (IOException e) {
+            catch (Exception e) {
                 if (!latestOnly) {
                     queue.enqueue(new EOFEvent(file, false));
                 }
                 LOGGER.warn("Error occurred while processing commit log " + file.getName(), e);
```
jpechane, line 128: This should be warning only when exception is not rethrown, otherwise it should be error.

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CommitLogProcessor.java
bingqinzhou: Nice catch! Correct codes accordingly :)

bingqinzhou: > Do we really want to use equality operator here? Also `DEFAULT_COMMIT_LOG_TRANSFER_CLASS` contians the name of the calss but the `toString` here is called the `class` itself.

Nice catch! Correct codes accordingly :)

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CommitLogProcessor.java
bingqinzhou: > This should be warning only when exception is not rethrown, otherwise it should be error.

I see, updated accordingly

jpechane: I think was not fully clear, my intent was something like

```
if (...) {
    LOGGER.error
    throw e;
}
LOGGER.warn
```

Also could you please rebase upon the lastest master? There was a change in the `debezium` repo breaking builds for this one.

bingqinzhou: > I think was not fully clear, my intent was something like
> 
> ```
> if (...) {
>     LOGGER.error
>     throw e;
> }
> LOGGER.warn
> ```
> 
> Also could you please rebase upon the lastest master? There was a change in the `debezium` repo breaking builds for this one.

Updated accordingly!

jpechane:
@bingqinzhou Applied, thanks!

PR #179, opened 2020-07-09
DBZ-137 log miner oracle contribution

AndreyIg:

PR #185, opened 2020-08-24
DBZ-137 log miner oracle contribution

Naros:
This replaces PR #179 that has been rebased against master.

Requires the following Pull Requests before merging:
https://github.com/debezium/oracle-vagrant-box/pull/12
https://github.com/debezium/debezium/pull/1709


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/jsqlparser/SimpleDmlParser.java
```diff
@@ -0,0 +1,287 @@
...
+import net.sf.jsqlparser.statement.Statement;
+import net.sf.jsqlparser.statement.delete.Delete;
+import net.sf.jsqlparser.statement.insert.Insert;
+import net.sf.jsqlparser.statement.update.Update;
+
+/**
+ * This class does parsing of simple DML: insert, update, delete.
+ * Log Miner supplies very simple syntax , that this parser should be sufficient to parse those.
+ * It does no support joins, merge, sub-selects and other complicated cases, which should be OK for Log Miner case
+ */
```
MilovdZee, line 51: Completely agree and this is why I use simple RegEx statements to get rid of jsqlparser. jsqlparser is very slow... But maybe you are using the fast parts...
For example `Regex("(?i)^UPDATE .* SET (.*) WHERE (.*)$")` and then split on AND

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
```diff
@@ -0,0 +1,134 @@
...
+    static String queryLogMinerContents(String schemaName, String logMinerUser, OracleDatabaseSchema schema) {
+        List<String> whiteListTableNames = schema.tableIds().stream().map(TableId::table).collect(Collectors.toList());
+
+        String sorting = "ORDER BY SCN";
+        return "SELECT SCN, SQL_REDO, OPERATION_CODE, TIMESTAMP, XID, CSF, TABLE_NAME, SEG_OWNER, OPERATION, USERNAME " +
+                " FROM " + LOGMNR_CONTENTS_VIEW + " WHERE  OPERATION_CODE in (1,2,3,5) " + // 5 - DDL
+                " AND SEG_OWNER = '" + schemaName.toUpperCase() + "' " +
+                buildTableInPredicate(whiteListTableNames) +
+                " AND SCN >= ? AND SCN < ? " +
+                " OR (OPERATION_CODE IN (5,7,34,36) AND USERNAME NOT IN ('SYS','SYSTEM','" + logMinerUser.toUpperCase() + "'))" + sorting; // todo username = schemaName?
```
MilovdZee, line 111: 34 as well? MISSING_SCN. Interesting.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
```diff
@@ -0,0 +1,263 @@
...
+                LOGGER.trace("current millis {}, db time {}", System.currentTimeMillis(), getTimeDifference(connection));
+                transactionalBufferMetrics.setTimeDifference(new AtomicLong(getTimeDifference(connection)));
+
+                if (!isContinuousMining && startScn < getFirstOnlineLogScn(connection)) {
+                    throw new RuntimeException("Online REDO LOG files don't contain the offset SCN. Clean offset and start over");
+                }
+
+                // 1. Configure Log Miner to mine online redo logs
+                setNlsSessionParameters(jdbcConnection);
+                checkSupplementalLogging(jdbcConnection, connectorConfig.getPdbName());
```
MilovdZee, line 134: This could be done once at startup and does not have to be done every time again.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
```diff
@@ -0,0 +1,134 @@
...
+    static final String OLDEST_FIRST_CHANGE = "SELECT MIN(FIRST_CHANGE#) FROM V$LOG";
+    static final String ALL_ONLINE_LOGS = "SELECT MIN(F.MEMBER) AS FILE_NAME, L.NEXT_CHANGE# AS NEXT_CHANGE, F.GROUP# " +
+            " FROM V$LOG L, V$LOGFILE F " +
+            " WHERE F.GROUP# = L.GROUP# AND L.NEXT_CHANGE# > 0 " +
+            " GROUP BY F.GROUP#, L.NEXT_CHANGE# ORDER BY 3";
+
+    static final String REDO_LOGS_STATUS = "SELECT F.MEMBER, R.STATUS FROM V$LOGFILE F, V$LOG R WHERE F.GROUP# = R.GROUP# ORDER BY 2";
+    static final String SWITCH_HISTORY_TOTAL_COUNT = "select 'total', count(1) from v$archived_log where first_time > trunc(sysdate)" +
+            " and dest_id = (select dest_id from V$ARCHIVE_DEST_STATUS where status='VALID' and type='LOCAL')";
+    static final String CURRENT_REDO_LOG_NAME = "select f.member from v$log log, v$logfile f  where log.group#=f.group# and log.status='CURRENT'";
```
MilovdZee, line 40: You only need one. If there is a group just take one of the files in the group.
```
        val CURRENT_REDO_FILENAME = """
            SELECT DISTINCT(first_value(lf.MEMBER) OVER (PARTITION BY lf.GROUP# ORDER BY lf.MEMBER)) AS FILENAME,
                   FIRST_CHANGE# AS FIRST,
                   NEXT_CHANGE# - 1 AS LAST
            FROM V$LOGFILE lf, V$LOG l
            WHERE lf.GROUP# = l.GROUP# AND l.STATUS = 'CURRENT'
        """.trimIndent()
```

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHelper.java
```diff
@@ -0,0 +1,431 @@
...
+     * @param connection connection to reuse
+     * @param metrics MBean accessible metrics
+     * @return full redo log file name(s), including path
+     * @throws SQLException if anything unexpected happens
+     */
+    static Set<String> getCurrentRedoLogFiles(Connection connection, LogMinerMetrics metrics) throws SQLException {
+        String checkQuery = SqlUtils.CURRENT_REDO_LOG_NAME;
+
+        Set<String> fileNames = new HashSet<>();
+        PreparedStatement st = connection.prepareStatement(checkQuery);
```
MilovdZee, line 175: try with resource

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
```diff
@@ -0,0 +1,263 @@
...
+                        // This is the way to mitigate PGA leak.
+                        // With one mining session it grows and maybe there is another way to flush PGA, but at this point we use new mining session
+                        endMining(connection);
+
+                        if (!isContinuousMining) {
+                            if (strategy == OracleConnectorConfig.LogMiningStrategy.CATALOG_IN_REDO) {
+                                buildDataDictionary(connection);
+                            }
+
+                            abandonOldTransactionsIfExist(connection);
```
MilovdZee, line 170: I put the miners in a cluster and store abandoned transactions in Hazelcast for later retrieval. This is a bit of shame to throw the transaction away. And what happens if the transaction is huge? Then the switch keeps happening before the transaction is handled. Or did you handle that situation somewhere else?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/jsqlparser/SimpleDmlParser.java
Naros: I understand its likely slow and the contributor has another implementation coming but we needed to get this rebased & integrated.  Once we have the parser code submitted, I'd be interested in hearing from you on performance if you have the bandwidth.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
Naros: @AndreyIg do you have any input on this?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
Naros: I agree.  We have some more changes coming from the contributor so I'll temporarily add a todo for this and once those other contributions are in, we can rework this.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
Naros: @AndreyIg any insight on this suggestion?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
Naros: From what I can gather from the code, I suspect the transaction is then incomplete in Kafka.  @AndreyIg is that accurate?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
```diff
@@ -0,0 +1,263 @@
...
+                while (context.isRunning()) {
+
+                    endScn = getEndScn(connection, startScn, logMinerMetrics);
+                    // LOGGER.trace("startScn: {}, endScn: {}", startScn, endScn);
+
+                    metronome = Metronome.sleeper(Duration.ofMillis(logMinerMetrics.getMillisecondToSleepBetweenMiningQuery()), clock);
+                    metronome.pause();
+
+                    Set<String> possibleNewCurrentLogFile = getCurrentRedoLogFiles(connection, logMinerMetrics);
+                    if (!currentRedoLogFiles.equals(possibleNewCurrentLogFile)) {
```
MilovdZee, line 159: If I understand correctly the check is only done before doing the work. So if the amount of records to process is huge the switch can happen during the processing. I always check the switch every 10 seconds.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerQueryResultProcessor.java
```diff
@@ -0,0 +1,245 @@
...
+     */
+    int processResult(ResultSet resultSet) {
+        int dmlCounter = 0;
+        int commitCounter = 0;
+        int rollbackCounter = 0;
+        Duration cumulativeCommitTime = Duration.ZERO;
+        Duration cumulativeParseTime = Duration.ZERO;
+        Duration cumulativeOtherTime = Duration.ZERO;
+        Instant startTime = Instant.now();
+        while (true) {
```
MilovdZee, line 84: In this loop the redo log switching should also be checked. Not for every record because that would be to expensive but I do it every 10 seconds. Assuming the switching of all redo files takes more then 10 seconds to go around and reuse the active redo log.

MilovdZee:
I see in TransactionBuffer that a BigDecimal is used for the scn. I would suggest to use a BigInteger.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerQueryResultProcessor.java
```diff
@@ -0,0 +1,245 @@
...
+
+                dmlEntry.setObjectOwner(segOwner);
+                dmlEntry.setSourceTime(changeTime);
+                dmlEntry.setTransactionId(txId);
+                dmlEntry.setObjectName(tableName);
+                dmlEntry.setScn(scn);
+
+                try {
+                    TableId tableId = RowMapper.getTableId(catalogName, resultSet);
+                    transactionalBuffer.registerCommitCallback(txId, scn, changeTime.toInstant(), redo_sql, (timestamp, smallestScn, commitScn, counter) -> {
```
MilovdZee, line 181: Why is there a commit callback for every record? With 100.000 records in a transaction this results in many callbacks... That probably all are very much the same. Why not just store a meta record at the start of a transaction (first occurrence of the idx) and the add only the sql to the list for that transaction?

MilovdZee: Also how do you handle very large transactions? I don't see optimizations for the memory usage.

Naros:
> I see in TransactionBuffer that a BigDecimal is used for the scn. I would suggest to use a BigInteger.

I have added https://issues.redhat.com/browse/DBZ-2457 as a follow-up to handle this.  I agree that we should use a `BigInteger` rather than a `BigDecimal`.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerQueryResultProcessor.java
Naros: > In this loop the redo log switching should also be checked. Not for every record because that would be to expensive but I do it every 10 seconds. Assuming the switching of all redo files takes more then 10 seconds to go around and reuse the active redo log.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
Naros: Added as a follow-up under https://issues.redhat.com/browse/DBZ-2457.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerQueryResultProcessor.java
Naros: Added as a follow-up under https://issues.redhat.com/browse/DBZ-2457.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerQueryResultProcessor.java
Naros: > Why is there a commit callback for every record? With 100.000 records in a transaction this results in many callbacks... That probably all are very much the same. Why not just store a meta record at the start of a transaction (first occurrence of the txId) and the add only the sql to the list for that transaction?

I most definitely agree that this could be reworked and far more memory friendly.  I've logged https://issues.redhat.com/browse/DBZ-2459 as a follow-up for this.

Naros: > Also how do you handle very large transactions? I don't see optimizations for the memory usage.

I assume you're referring to the situation where the maps/sets grow but any excess memory that may be consumed by large or long running transactions is never recovered once they've been committed?

MilovdZee: In a high traffic environment (4000 mutations/sec) with large transactions of up to 100.000 in size, keeping a list with mutations and throwing the list away after use will introduce a lot of GC. So I think we have to be very much aware of optimizing the transaction storage and probably even find a way that GC does not get involved. So maybe some kind of custom memory management... I don't know how to do that yet. I'm thinking in the direction of using a dictionary with the most common words and referencing them. Column names for example probably are very stable and duplicated in every mutation.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/jsqlparser/SimpleDmlParser.java
MilovdZee: I did not do a thorough and detailed performance test but when I used jsqlParser the CPU went to 100% all the time and parsing was slow (I reached about 400 msg/sec). Then I changed to RegEx and now the CPU stays relatively low and am at about 4000 msg/sec.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
AndreyIg: By doing this I was intended to address a few non-standard situations. There a re more on todos

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
AndreyIg: sure

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
AndreyIg: it would be more than one in RAC environment

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
AndreyIg: OLTP supposed to use short transactions. If a transaction lasts hours, there is something wrong in the application design in my opinion. Without abandoning mechanism, we might end up with full re-snapshot after restarting the connector.
Huge transaction should not be a problem,as long as it is fits let say in one hour

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
AndreyIg: I think it is also good idea, but checking before mining works too

MilovdZee: Only if the mine does not take very long. If the connector has been down for some time the amount of files to mine if increasing and so is the time it takes to handle them all. Chances are that during the mining the redo log rotates. At least, that is what I see in my environment. So I check every 10 seconds.

AndreyIg:
As a general comment, I implemented a few new things:
- mining history recording (in performance testing)
- RAC support, tested
- mining archives for continuous_mine=false
I'm waiting for PR to be merged to rebase my branch and submit new PR

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
AndreyIg: This is a valid concern, however after a switch, the log still has all the contents to fetch. So, rotation should not be a problem unless this is the last redo group to rotate and has to be cleaned and reused. We recommend 7 groups (0.5G each file), so assuming each group will last for 10 minutes, we a good for 1 hour. Anyway checking every 10 seconds should be more secure.

MilovdZee: Ok. That is different from the setup I'm working with. Here sometimes the logs rotate within 10 seconds and we only have three files. This is our TST environment. In PRD we have 5 groups of 1G each that last a lot longer each. Still I prefer to check every couple of seconds. Costs are very low and safety gain is high.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
MilovdZee: Ok.

MilovdZee: I can't find that the archive_logs are used. This is the query I use to get all log files. Archived and online combined.
```
            SELECT *
            FROM (
                 SELECT DISTINCT(first_value(lf.MEMBER) OVER (PARTITION BY lf.GROUP# ORDER BY lf.MEMBER)) AS FILENAME,
                        FIRST_CHANGE# AS FIRST,
                        NEXT_CHANGE# - 1 AS LAST
                 FROM V$LOGFILE lf, V$LOG l
                 WHERE lf.GROUP# = l.GROUP# AND l.STATUS = 'CURRENT'
                 UNION
                 SELECT NAME AS FILENAME, 
                        FIRST_CHANGE# AS FIRST, 
                        NEXT_CHANGE# - 1 AS LAST
                 FROM V$ARCHIVED_LOG
                 WHERE DELETED = 'NO' AND STANDBY_DEST = 'NO'
             )
            WHERE ? <= LAST
```

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/jsqlparser/SimpleDmlParser.java
AndreyIg: Having a cheap parser would be great.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
AndreyIg: Nice query. You are filtering out duplication later, right?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/jsqlparser/SimpleDmlParser.java
MilovdZee: First thing I do is remove the newlines and replace them with a placeholder for later restoring.
```
private val newlineInFieldPattern = Regex("('[^'\n]*)\r?\n(?!(([^']*'){2})*[^']*\$)")
val cleanedSqlRedo = sqlRedo.replace(newlineInFieldPattern, "$1$NEWLINE_STRING").replace("\n", " ")
```
Then for the different types of operations different regexes:
```
val regex = Regex("(?i)^INSERT INTO [^(]*\\((.*)\\) VALUES \\((.*)\\)$")
val setAndWhereRegEx = Regex("(?i)^UPDATE .* SET (.*) WHERE (.*)$")
val regex = Regex("(?i)^DELETE FROM .* WHERE (.*)$")
```

This separates the before and after values and groups them. Some simple splitting on 'AND' and fixing words like TIMESTAMP, IS NULL, etc

```
private val quotePattern = Regex("^[\\s]*['\"](.*)['\"][\\s]*$")
private val newlineInFieldPattern = Regex("('[^'\n]*)\r?\n(?!(([^']*'){2})*[^']*\$)")
val cleaned = str
                    .replace(timestampPattern, "") // remove TIMESTAMP word
                    .replace(quotePattern, "$1") // remove starting and trailing ' and "
                    .replace(NEWLINE_STRING, "\n")
                    .trim() // remove any surrounding spaces
```

MilovdZee:
```
@Service
class ResultSetService(
        @Value("\${app.source.brand}") val brand: String,
        @Value("\${app.source.domain}") val domain: String,
        @Value("\${app.source.datastore}") val datastore: String,
        @Value("\${app.source.timezone}") val timezone: String
) {
    companion object {
        private const val NULL_STRING = "NULL"
        private const val NEWLINE_STRING = "#NL#"

        // Match 'comma' but only if not enclosed in quotes
        val csvPattern = Regex(",(?=(?:[^']*'[^']*')*[^']*$)")

        private val timestampPattern = Regex("^[\\s]*TIMESTAMP[\\s]+")
        private val quotePattern = Regex("^[\\s]*['\"](.*)['\"][\\s]*$")
        private val newlineInFieldPattern = Regex("('[^'\n]*)\r?\n(?!(([^']*'){2})*[^']*\$)")

        fun cleanString(str: String): String {
            val timestampFound = timestampPattern.containsMatchIn(str)

            val cleaned = str
                    .replace(timestampPattern, "") // remove TIMESTAMP word
                    .replace(quotePattern, "$1") // remove starting and trailing ' and "
                    .replace(NEWLINE_STRING, "\n")
                    .trim() // remove any surrounding spaces

            // If it is a TIMESTAMP replace it with the number of milliseconds since epoch UTC
            return if (timestampFound)
                Timestamp.valueOf(cleaned).toInstant().toEpochMilli().toString()
            else
                cleaned
        }
    }

    fun getSource(dataSet: String): Source {
        return Source(
                brand = brand,
                domain = domain,
                datastore = datastore,
                dataset = dataSet,
                schema = "$dataSet-v1 schema",
                timezone = timezone
        )
    }

    fun mapResultToData(operation: SqlOperation, sqlRedo: String): TransactionEntry {
        val cleanedSqlRedo = sqlRedo.replace(newlineInFieldPattern, "$1$NEWLINE_STRING").replace("\n", " ")

        val (beforeValues, afterValues) = when (operation) {
            SqlOperation.INSERT -> handleInsert(cleanedSqlRedo)
            SqlOperation.UPDATE -> handleUpdate(cleanedSqlRedo)
            SqlOperation.DELETE -> handleDelete(cleanedSqlRedo)
            SqlOperation.SELECT_FOR_UPDATE -> error("$operation not implemented yet!")
            SqlOperation.COMMIT -> error("$operation should not be reachable! Fix the code!")
            SqlOperation.ROLLBACK -> error("$operation should not be reachable! Fix the code!")
            SqlOperation.START -> error("$operation should not be reachable! Fix the code!")
        }

        val fields = beforeValues.keys.union(afterValues.keys).toTypedArray()
        val beforeList = fields.map { field -> beforeValues[field] }.toTypedArray()
        val afterList = fields.map {field ->
            val beforeValue = beforeValues[field]
            val afterValue = afterValues[field]
            if(afterValue == beforeValue) beforeValue else afterValue
        }.toTypedArray()

        return TransactionEntry(
                operation = operation,
                fields = fields,
                before = beforeList,
                after = afterList
        )
    }

    private fun handleInsert(sqlRedo: String): Pair<Map<String, String>, Map<String, String>> {
        // Always in the form 'INSERT INTO "<schema>"."<table>"("<field_1>", ..., "<field_n>") VALUES (<value_1>, ..., <value_n>)'
        val regex = Regex("(?i)^INSERT INTO [^(]*\\((.*)\\) VALUES \\((.*)\\)$")
        val matches = regex.find(sqlRedo)!!
        val fieldsGroup = matches.groups[1]!!.value
        val valuesGroup = matches.groups[2]!!.value

        // Get the fields and the values
        val fields = fieldsGroup
                .split(csvPattern)
                .map { fieldName -> cleanString(fieldName) }
        val values = valuesGroup
                .split(csvPattern)
                .map { value -> cleanString(value) }

        // Create the map with field -> value pairs
        val afterValues = fields
                .mapIndexed { index, field -> field to values[index] }
                .toMap()

        val beforeValues = mapOf<String, String>()
        return Pair(beforeValues, afterValues)
    }

    private fun handleUpdate(sqlRedo: String): Pair<Map<String, String>, Map<String, String>> {
        // Almost always in the form:
        //   UPDATE "<schema>"."<table>"
        //   SET "<field_1>" = "<value_1>", ..., "<field_n>" = "<value_n>"
        //   WHERE "<field_1>" = "<value_1>" AND ... AND "<field_n>" = "<value_n>"
        val setAndWhereRegEx = Regex("(?i)^UPDATE .* SET (.*) WHERE (.*)$")
        val matchesSetAndWhere = setAndWhereRegEx.find(sqlRedo)

        var matchesSetWithoutWhere: MatchResult? = null
        if (matchesSetAndWhere == null) {
            // Or else might be without the where:
            //   UPDATE "<schema>"."<table>"
            //   SET "<field_1>" = "<value_1>", ..., "<field_n>" = "<value_n>"
            val setWithoutWhereRegEx = Regex("(?i)^UPDATE .* SET (.*)$")
            matchesSetWithoutWhere = setWithoutWhereRegEx.find(sqlRedo) ?: error("handleUpdate: unable to handle $sqlRedo")
        }
        val setStatementsGroup = matchesSetAndWhere?.groups?.get(1)?.value ?: matchesSetWithoutWhere!!.groups[1]!!.value
        val whereStatementGroup = matchesSetAndWhere?.groups?.get(2)?.value ?: ""

        // Extract the important parts of the statement
        val setStatements = setStatementsGroup
                .split(csvPattern)
                .map { pair -> pair.split(" = ").let { Pair(cleanString(it[0]), cleanString(it[1])) } }
        val whereStatements = matchesSetAndWhere?.let {
            whereStatementGroup
                    .split(" AND ", ignoreCase = true)
                    .map { pair -> pair.split(" = ", " IS ").let { Pair(cleanString(it[0]), cleanString(it[1])) } }
        } ?: emptyList()

        // Before values are just the WHERE values. Logminer sends all field in the WHERE
        val beforeValues = whereStatements.toMap().toMutableMap()

        // Use the beforeValues as start for the afterValues and overwrite with changes
        val afterValues = beforeValues.toMutableMap()
        setStatements.forEach { (field, value) -> afterValues[field] = value }

        // If both before and after field contains 'null' then just ignore the field
        beforeValues
                .filter { (field, value) -> value == NULL_STRING && (afterValues[field]?.let { it == NULL_STRING } ?: true) }
                .forEach { (field, _) ->
                    beforeValues.remove(field)
                    afterValues.remove(field)
                }

        return Pair(beforeValues, afterValues)
    }

    private fun handleDelete(sqlRedo: String): Pair<Map<String, String>, Map<String, String>> {
        // Always in the form 'DELETE FROM "<schema>"."<table>" WHERE "<field_1>" = "<value_1>" AND ... AND "<field_n>" = "<value_n>"'
        val regex = Regex("(?i)^DELETE FROM .* WHERE (.*)$")
        val matches = regex.find(sqlRedo)!!
        val whereStatementGroup = matches.groups[1]!!.value

        val whereStatements = whereStatementGroup
                .split(" AND ", ignoreCase = true)
                .map { fieldValuePair -> fieldValuePair.split(" = ", " IS ").let { Pair(cleanString(it[0]), cleanString(it[1])) } }

        // Before values are just the WHERE values. Logminer sends all field in the WHERE
        val beforeValues = whereStatements.toMap()

        val afterValues = mapOf<String, String>()
        return Pair(beforeValues, afterValues)
    }
}
```

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
MilovdZee: I did not see any duplicates up to today actually. It only happened once. Was that because of rotation and the having some overlapping states? But indeed. I have to filter for duplicates and I'll do that in this query actually.

MilovdZee: `STANDBY_DEST = 'NO'` seems to be needed. Otherwise the filenames can have weird values. I'm no expert on Oracle so this for me is trial and error. Fail fast, fix forward :)

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -91,6 +105,55 @@
...
+            .withDescription("If true, CONTINUOUS_MINE option will be added to the log mining session. This will manage log files switches seamlessly.");
+
+    public static final Field SNAPSHOT_ENHANCEMENT_TOKEN = Field.create("snapshot.enhance.predicate.scn")
+            .withDisplayName("A string to replace on snapshot predicate enhancement")
+            .withType(Type.STRING)
+            .withWidth(Width.MEDIUM)
+            .withImportance(Importance.HIGH)
+            .withDescription("A token to replace on snapshot predicate template");
+
+    public static final Field DRIVER_TYPE = Field.create("database.driver.type")
```
jpechane, line 149: If there is hard dependency between `CONNECTOR_ADAPTER` and `DRIVER_TYPE` then it should either be controlled by a single config param or at least there should be a validator rejecting invalid combinations.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -123,35 +193,60 @@
...
         this.xoutServerName = config.getString(XSTREAM_SERVER_NAME);
         this.snapshotMode = SnapshotMode.parse(config.getString(SNAPSHOT_MODE));
         this.tablenameCaseInsensitive = config.getBoolean(TABLENAME_CASE_INSENSITIVE);
         this.oracleVersion = OracleVersion.parse(config.getString(ORACLE_VERSION));
+        this.schemaName = setUpperCase(config.getString(SCHEMA_NAME));
+        String blacklistedColumns = setUpperCase(config.getString(RelationalDatabaseConnectorConfig.COLUMN_BLACKLIST));
+        this.columnFilter = getColumnNameFilter(blacklistedColumns);
+    }
+
+    private String setUpperCase(String property) {
```
jpechane, line 213: Maybe rename to `toUpperCase`?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -123,35 +193,60 @@
...
+        property = property == null ? null : property.toUpperCase();
```
jpechane, line 214: Can be changed to a one-line method

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -180,19 +275,36 @@ public OracleVersion getOracleVersion() {
...
             protected boolean isPositionAtOrBefore(Document recorded, Document desired) {
-                final LcrPosition recordedPosition = LcrPosition.valueOf(recorded.getString(SourceInfo.LCR_POSITION_KEY));
-                final LcrPosition desiredPosition = LcrPosition.valueOf(desired.getString(SourceInfo.LCR_POSITION_KEY));
-                final Long recordedScn = recordedPosition != null ? recordedPosition.getScn() : recorded.getLong(SourceInfo.SCN_KEY);
-                final Long desiredScn = desiredPosition != null ? desiredPosition.getScn() : desired.getLong(SourceInfo.SCN_KEY);
-
-                return (recordedPosition != null && desiredPosition != null)
-                        ? recordedPosition.compareTo(desiredPosition) < 1
-                        : recordedScn.compareTo(desiredScn) < 1;
+                Long recordedScn;
```
jpechane, line 291: Could be as a follow-up introduces a Scn domain type?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -91,6 +105,55 @@
...
             .withValidation(CommonConnectorConfig::validateServerNameIsDifferentFromHistoryTopicName);
 
+    public static final Field CONNECTOR_ADAPTER = Field.create(DATABASE_CONFIG_PREFIX + "connection.adapter")
+            .withDisplayName("Connector adapter")
+            .withEnum(ConnectorAdapter.class, ConnectorAdapter.XSTREAM)
+            .withWidth(Width.MEDIUM)
+            .withImportance(Importance.HIGH)
+            .withDescription("There are two adapters: XStream and LogMiner.");
+
+    public static final Field SNAPSHOT_SKIP_LOCKS = Field.create("snapshot.skip.locks")
```
jpechane, line 115: How about using snapshot.locking.mode like we have for MySQL?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorTask.java
```diff
@@ -51,11 +51,15 @@ public ChangeEventSourceCoordinator start(Configuration config) {
         SchemaNameAdjuster schemaNameAdjuster = SchemaNameAdjuster.create(LOGGER);
 
         Configuration jdbcConfig = config.subset("database.", true);
-        jdbcConnection = new OracleConnection(jdbcConfig, new OracleConnectionFactory());
+        jdbcConnection = new OracleConnection(jdbcConfig, new OracleConnectionFactory(), () -> getClass().getClassLoader());
         this.schema = new OracleDatabaseSchema(connectorConfig, schemaNameAdjuster, topicSelector, jdbcConnection);
         this.schema.initializeStorage();
 
-        OffsetContext previousOffset = getPreviousOffset(new OracleOffsetContext.Loader(connectorConfig));
+        String adapterString = config.getString("connection.adapter");
```
jpechane, line 58: Should not use string but field name

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleOffsetContext.java
```diff
@@ -235,12 +254,20 @@ public Loader(OracleConnectorConfig connectorConfig) {
 
         @Override
         public OffsetContext load(Map<String, ?> offset) {
-            LcrPosition lcrPosition = LcrPosition.valueOf((String) offset.get(SourceInfo.LCR_POSITION_KEY));
-            Long scn = lcrPosition != null ? lcrPosition.getScn() : (Long) offset.get(SourceInfo.SCN_KEY);
             boolean snapshot = Boolean.TRUE.equals(offset.get(SourceInfo.SNAPSHOT_KEY));
             boolean snapshotCompleted = Boolean.TRUE.equals(offset.get(SNAPSHOT_COMPLETED_KEY));
+            Long scn;
+            if (adapter == OracleConnectorConfig.ConnectorAdapter.LOG_MINER) {
```
jpechane, line 260: Can this and other instances depending on the connector adapter be extracted directly into a s separate connector adapter startegy?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
```diff
@@ -28,7 +29,7 @@
  */
 class LcrEventHandler implements XStreamLCRCallbackHandler {
 
-    private static final Logger LOGGER = LoggerFactory.getLogger(OracleStreamingChangeEventSource.class);
+    private static final Logger LOGGER = LoggerFactory.getLogger(XstreamStreamingChangeEventSource.class);
```
jpechane, line 32: Logger name is incorrect

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
Naros: I've added this as [DBZ-2518](https://issues.redhat.com/browse/DBZ-2518)

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
Naros: Done

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
Naros: Done

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
Naros: I went ahead and removed this configuration property and rolled this behavior into the enum class with `#getDriverType()`.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
Naros: Fixed

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/jsqlparser/SimpleDmlParser.java
MilovdZee: I rewrote the above code to get rid of the RegEx parts. They were introducing a lot of garbage and were not fast enough. I now do a sqlRedo.forEach { c -> ... } and so parse the string only once. This is ten times as fast. And even fixes the problem of having ' WHERE ' in your values that failed with the RegEx.
If wanted I can share the code... As inspiration. Still needs to be thoroughly tested though.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerQueryResultProcessor.java
MilovdZee: I got rid of most of the GC issues by using a charArray and reusing this for string storage. I think that something like this also has to be done in this connector. Otherwise it will never be able to handle big load with large transactions.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorTask.java
gunnarmorling: Indeed, that seems superfluous. Just get it, as done in the null case on the next line.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDatabaseSchema.java
```diff
@@ -42,6 +43,10 @@ public OracleDatabaseSchema(OracleConnectorConfig connectorConfig, SchemaNameAdj
                 connectorConfig.getKeyMapper());
     }
 
+    public Tables getTables() {
```
gunnarmorling, line 46: Is that needed, doesn't add any value it seems?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
```diff
@@ -195,14 +205,40 @@ protected void readTableStructure(ChangeEventSourceContext sourceContext, Relati
                 throw new InterruptedException("Interrupted while reading structure of schema " + schema);
             }
 
+            // todo: DBZ-137 the new readSchemaForCapturedTables seems to cause failures.
```
gunnarmorling, line 208: Log a Jira issue?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
```diff
@@ -195,14 +205,40 @@ protected void readTableStructure(ChangeEventSourceContext sourceContext, Relati
...
                     snapshotContext.catalogName,
                     schema,
                     connectorConfig.getTableFilters().dataCollectionFilter(),
                     null,
                     false);
+            // }
+        }
+    }
+
+    @Override
```
gunnarmorling, line 232: I'd remove that here and revisit this more holistically in https://issues.redhat.com/browse/DBZ-2525.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
```diff
@@ -231,11 +267,44 @@ protected SchemaChangeEvent getCreateTableEvent(RelationalSnapshotContext snapsh
...
 
+    /**
+     * This is to build "whitelisted" column list
+     * @param blackListColumnStr comma separated columns blacklist
+     * @param table the table
+     * @return column list for select
+     */
+    public static String buildSelectColumns(String blackListColumnStr, Table table) {
+        String columnsToSelect = "*";
+        if (blackListColumnStr != null && blackListColumnStr.trim().length() > 0
```
gunnarmorling, line 285: Shouldn't use "blacklist", but see above.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSourceInfoStructMaker.java
```diff
@@ -23,6 +23,7 @@ public OracleSourceInfoStructMaker(String connector, String version, CommonConne
                 .field(SourceInfo.TABLE_NAME_KEY, Schema.STRING_SCHEMA)
                 .field(SourceInfo.TXID_KEY, Schema.OPTIONAL_STRING_SCHEMA)
                 .field(SourceInfo.SCN_KEY, Schema.OPTIONAL_INT64_SCHEMA)
+                .field(SourceInfo.COMMIT_SCN_KEY, Schema.OPTIONAL_INT64_SCHEMA)
```
gunnarmorling, line 26: Needs reflection in doc examples?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
```diff
@@ -92,12 +101,17 @@ private SchemaBuilder getNumericSchema(Column column) {
             if (scale <= 0) {
                 int width = column.length() - scale;
 
+                if (scale == 0 && column.length() == 1) {
```
gunnarmorling, line 104: Wouldn't this cause issue with something like NUMERIC(1)? Feeling a bit concerned about this one.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
```diff
@@ -149,11 +169,23 @@ public ValueConverter converter(Column column, Field fieldDefn) {
...
+        }
+        return convertVariableScale(column, fieldDefn, data);
+    }
+
     private ValueConverter getNumericConverter(Column column, Field fieldDefn) {
         if (column.scale().isPresent()) {
             Integer scale = column.scale().get();
 
             if (scale <= 0) {
+                // Boolean represtented as Number(1,0)
```
gunnarmorling, line 184: See above, it seems we couldn't export NUMBER(1, 0) as number any longer with that?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/antlr/OracleDmlParser.java
```diff
@@ -0,0 +1,95 @@
...
+import io.debezium.ddl.parser.oracle.generated.PlSqlParser;
+import io.debezium.relational.SystemVariables;
+import io.debezium.relational.Tables;
+
+/**
+ * This is the main Oracle Antlr DML parser
+ */
+public class OracleDmlParser extends AntlrDdlParser<PlSqlLexer, PlSqlParser> {
+
+    private LogMinerDmlEntry dmlEntry;
```
gunnarmorling, line 28: Can be made (some/all) final?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/jsqlparser/SimpleDmlParser.java
gunnarmorling: Thanks a lot, @MilovdZee, that's great input!

@Naros, could you capture this in a follow-up Jira issue? Indeed we might get away with just a regexp if the DML structure is so uniform and simple.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/jsqlparser/SimpleDmlParser.java
```diff
@@ -0,0 +1,287 @@
...
+public class SimpleDmlParser {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(SimpleDmlParser.class);
+    protected String catalogName;
```
gunnarmorling, line 55: Can be made some final? Also this could be using a `TableId` here?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHelper.java
```diff
@@ -0,0 +1,428 @@
...
+
+        return isNextScnCloseToDbCurrent ? currentScn : startScn + miningDiapason;
+    }
+
+    /**
+     * Calculate time difference between database and connector timers. It could be negative if DB time is ahead.
+     * @param connection connection
+     * @return difference in milliseconds
+     */
+    static long getTimeDifference(Connection connection) throws SQLException {
```
gunnarmorling, line 130: Could be Period or Duration.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerMetrics.java
```diff
@@ -0,0 +1,218 @@
...
+import io.debezium.connector.common.CdcSourceTaskContext;
+import io.debezium.metrics.Metrics;
+
+/**
+ * This class contains methods to be exposed via MBean server
+ *
+ */
+@ThreadSafe
+public class LogMinerMetrics extends Metrics implements LogMinerMetricsMXBean {
+    private AtomicLong currentScn = new AtomicLong();
```
gunnarmorling, line 26: Can be final? Also metrics need discussion in docs?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerMetrics.java
```diff
@@ -0,0 +1,218 @@
...
+    private AtomicReference<Duration> lastLogMinerQueryDuration = new AtomicReference<>();
+    private AtomicReference<Duration> averageLogMinerQueryDuration = new AtomicReference<>();
+    private AtomicInteger logMinerQueryCount = new AtomicInteger();
+    private AtomicReference<Duration> lastProcessedCapturedBatchDuration = new AtomicReference<>();
+    private AtomicInteger processedCapturedBatchCount = new AtomicInteger();
+    private AtomicReference<Duration> averageProcessedCapturedBatchDuration = new AtomicReference<>();
+    private AtomicInteger batchSize = new AtomicInteger();
+    private AtomicInteger millisecondToSleepBetweenMiningQuery = new AtomicInteger();
+
+    private final int MAX_SLEEP_TIME = 3_000;
```
gunnarmorling, line 40: Shold be static and go above.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerQueryResultProcessor.java
gunnarmorling: @MilovdZee, could you add this as a comment to the Jira issue created by @Naros and linked above?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
gunnarmorling: @Naros, can you log a follow-up?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
gunnarmorling: @Naros follow-up for this too?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
gunnarmorling: Needs tracking, @Naros?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/TransactionalBufferMetrics.java
```diff
@@ -0,0 +1,284 @@
...
+import io.debezium.annotation.ThreadSafe;
+import io.debezium.connector.common.CdcSourceTaskContext;
+import io.debezium.metrics.Metrics;
+
+/**
+ * This class contains MBean methods
+ */
+@ThreadSafe
+public class TransactionalBufferMetrics extends Metrics implements TransactionalBufferMetricsMXBean {
+    private AtomicLong oldestScn = new AtomicLong();
```
gunnarmorling, line 25: Can all be final?

gunnarmorling (commented):
@Naros some comments inline; admittedly mostly of cosmetic nature, I didn't dive into the full implementation details. Still it seems when addressing those and making sure the discussion points/todos are captured in Jira, we might merge it, clearly marking it as  work-in-progress and "don't use in prod"? WDYT? @jpechane, your feedback would be great, too?

@AndreyIg Thanks a lot for this huge piece of work!

@MilovdZee Thanks for reviewing and commenting! It sounds as if you're working on something similar? Is this open-source somewhere? If you got specific improvements, any PR from your side would be very welcomed!

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/jsqlparser/SimpleDmlParser.java
MilovdZee: For me the RegEx was not fast enough. And keep in minds that it was also not fail safe. But it was about the concept. For example my RegEx failed if there was a " WHERE " between quotes in one of the values.
I now do it even more manual and am able to handle up to 25000 UPDATES a second on my laptop. For INSERT this is even close to a 100.000. The only big problem is that the code is quite unreadable...

```
@Service
class ResultSetService(
        @Value("\${app.source.brand}") val brand: String,
        @Value("\${app.source.domain}") val domain: String,
        @Value("\${app.source.datastore}") val datastore: String,
        @Value("\${app.source.timezone}") val timezone: String
) {
    companion object {
        private const val MAX_STRING_SIZE = 10000
        private const val NULL_STRING = "NULL"
        private const val NOT_IN_QUOTES = '-'

        private val setArray = "SET".toCharArray()
        private val whereArray = "WHERE".toCharArray()
    }

    // Buffer used while parsing the sql statements
    // The result of using a shared buffer is that the service is not thread-safe...
    private val string = CharArray(MAX_STRING_SIZE)
    private var stringIndex = 0

    fun getSource(dataSet: String): Source {
        return Source(
                brand = brand,
                domain = domain,
                datastore = datastore,
                dataset = dataSet,
                schema = "$dataSet-v1 schema",
                timezone = timezone
        )
    }

    fun mapResultToData(operation: SqlOperation, sqlRedo: String): TransactionEntry {
        val (beforeValues, afterValues) = when (operation) {
            SqlOperation.INSERT -> handleInsert(sqlRedo)
            SqlOperation.UPDATE -> handleUpdate(sqlRedo)
            SqlOperation.DELETE -> handleDelete(sqlRedo)
            SqlOperation.SELECT_FOR_UPDATE -> error("$operation not implemented yet!")
            SqlOperation.COMMIT -> error("$operation should not be reachable! Fix the code!")
            SqlOperation.ROLLBACK -> error("$operation should not be reachable! Fix the code!")
            SqlOperation.START -> error("$operation should not be reachable! Fix the code!")
        }

        return TransactionEntry(
                operation = operation,
                before = beforeValues,
                after = afterValues
        )
    }

    private fun handleInsert(sqlRedo: String): Pair<Map<String, String>, Map<String, String>> {
        // Always in the form 'INSERT INTO "<schema>"."<table>"("<field_1>", ..., "<field_n>") VALUES (<value_1>, ..., <value_n>)'
        // So:
        // - two blocks between braces. First block contains the fields and the second block contains the values
        // - Steps:
        //   - find the first opening brace
        //   - read comma separated fields until the closing brace
        //   - find next opening brace
        //   - read comma separated values until the closing brace
        //   - combine fields and values into result

        var inQuotes = false
        var inBrackets = false
        var inBlock = false
        var timestamp = false
        var state = InsertBlockSequence.FIELDS

        val fields = mutableListOf<String>()
        val values = mutableListOf<String>()

        sqlRedo.forEach { c ->
            if (c == '"' || c == '\'') {
                // Remove quotes from the string and keep track of if we are inside quotes
                inQuotes = !inQuotes
            } else if (!inQuotes && c == '(') {
                // remove the brackets from the string keep track of if we are inside brackets
                inBrackets = true
            } else if (inBrackets) {
                if (inBlock && !inQuotes) {
                    if (c == ',' || c == ')' || c == ' ') {
                        // Handle end of block
                        inBlock = false
                        when (state) {
                            InsertBlockSequence.FIELDS -> {
                                fields.add(String(string, 0, stringIndex))
                                stringIndex = 0
                            }
                            InsertBlockSequence.VALUES -> {
                                val value = if (timestamp)
                                    Timestamp.valueOf(String(string, 0, stringIndex)).toInstant().toEpochMilli().toString()
                                else
                                    String(string, 0, stringIndex)
                                timestamp = value == "TIMESTAMP"
                                if (!timestamp) {
                                    values.add(value)
                                }
                                stringIndex = 0
                            }
                        }

                        // Closing bracket, so go to the next state
                        if (c == ')') {
                            state = InsertBlockSequence.VALUES
                            inBrackets = false
                        }
                    }
                } else if (c != ' ' && c != '\n') {
                    // It is not a space so must be inside some kind of block
                    inBlock = true
                }

                // Everything within quotes is always part of the block
                if (inBlock || inQuotes) {
                    string[stringIndex++] = c
                    if(stringIndex == MAX_STRING_SIZE) error("handleInsert: string to long to fit in buffer ($sqlRedo)")
                }
            }
        }

        val beforeValues = mapOf<String, String>()
        val afterValues = fields.mapIndexed { index, field -> field to values[index] }.toMap()
        return Pair(beforeValues, afterValues)
    }

    private fun handleUpdate(sqlRedo: String): Pair<Map<String, String>, Map<String, String>> {
        // Almost always in the form:
        //   UPDATE "<schema>"."<table>"
        //   SET "<field_1>" = '<value_1>', ..., "<field_n>" = '<value_n>'
        //   WHERE "<field_1>" = '<value_1>' AND ... AND "<field_n>" = '<value_n>'
        //
        // Notes:
        // - 'value' could be "NULL" without quotes
        // - 'value' could be "TIMESTAMP ' 2020-04-23 00:00:00'"
        // - "=" could also be "IS"

        val (beforeMap, afterMap) = extractBeforeAndAfterMap(sqlRedo, false)

        // Add missing fields to the afterMap
        beforeMap.entries
                .filter { (_, value) -> value != NULL_STRING }
                .filter { (field, _) -> !afterMap.containsKey(field) }
                .forEach { (field, value) -> afterMap[field] = value }

        // If both before and after field contains 'null' then just ignore the field
        beforeMap
                .filter { (field, value) -> value == NULL_STRING && (afterMap[field]?.let { it == NULL_STRING } ?: true) }
                .forEach { (field, _) ->
                    beforeMap.remove(field)
                    afterMap.remove(field)
                }

        return Pair(beforeMap, afterMap)
    }

    private fun handleDelete(sqlRedo: String): Pair<Map<String, String>, Map<String, String>> {
        // Always in the form 'DELETE FROM "<schema>"."<table>" WHERE "<field_1>" = "<value_1>" AND ... AND "<field_n>" = "<value_n>"'
        return extractBeforeAndAfterMap(sqlRedo, true)
    }

    private fun extractBeforeAndAfterMap(sqlRedo: String, skipSet: Boolean): Pair<MutableMap<String, String>, MutableMap<String, String>> {
        var setIndex = 0
        var whereIndex = 0
        var setSeen = false
        var whereSeen = false

        val afterMap = mutableMapOf<String, String>()
        val beforeMap = mutableMapOf<String, String>()

        var timestamp = false
        var inQuotesChar = NOT_IN_QUOTES
        var key = ""
        var state = UpdateBlockSequence.KEY // Start by looking for a key
        var inBlock = false
        val sqlRedoLength = sqlRedo.length
        for (charIndex in 0..sqlRedoLength) {
            val c = if (charIndex == sqlRedoLength) '\n' else sqlRedo[charIndex]

            // Search for SET
            if (!setSeen && !skipSet) {
                if (setArray[setIndex] == c) {
                    setIndex++
                    if (setIndex == setArray.size) {
                        setSeen = true
                    }
                } else setIndex = 0
            } else {
                if (c == '"' || c == '\'') {
                    // Keep track of if we are inside quotes
                    if (inQuotesChar == NOT_IN_QUOTES) {
                        // Start of quoted part
                        inQuotesChar = c
                    } else if (inQuotesChar == c) {
                        // End of quoted part
                        inQuotesChar = NOT_IN_QUOTES
                    }
                }

                // Search for WHERE
                if (!whereSeen) {
                    if (inQuotesChar == NOT_IN_QUOTES && whereArray[whereIndex] == c) {
                        whereIndex++
                        if (whereIndex == whereArray.size) {
                            whereSeen = true
                            continue // handle the next character
                        }
                    } else whereIndex = 0
                }

                if (whereSeen || !skipSet) {
                    if (inBlock && inQuotesChar == NOT_IN_QUOTES) {
                        // Handle end of block
                        if (c == ',' || c == '=' || c == ' ' || c == '\n' || c == '"' || c == '\'') {
                            inBlock = false
                            when (state) {
                                UpdateBlockSequence.KEY -> {
                                    key = String(string, 0, stringIndex)
                                    stringIndex = 0
                                    state = if (c == ' ' || c == '\n' || c == '"' || c == '\'') UpdateBlockSequence.SEPERATOR_AFTER_KEY else UpdateBlockSequence.VALUE
                                }
                                UpdateBlockSequence.SEPERATOR_AFTER_KEY -> {
                                    state = UpdateBlockSequence.VALUE
                                }
                                UpdateBlockSequence.VALUE -> {
                                    val value = if (timestamp)
                                        Timestamp.valueOf(String(string, 0, stringIndex)).toInstant().toEpochMilli().toString()
                                    else
                                        String(string, 0, stringIndex)
                                    timestamp = value == "TIMESTAMP"
                                    if (!timestamp) {
                                        if (!whereSeen) afterMap[key] = value else beforeMap[key] = value
                                        state = if (c == ' ' || c == '\n' || c == '"' || c == '\'') UpdateBlockSequence.SEPARATOR_AFTER_VALUE else UpdateBlockSequence.KEY
                                    }
                                    stringIndex = 0
                                }
                                UpdateBlockSequence.SEPARATOR_AFTER_VALUE -> {
                                    state = UpdateBlockSequence.KEY
                                }
                            }
                        }
                    } else if (inQuotesChar != NOT_IN_QUOTES || (c != ' ' && c != '\n')) {
                        // It is not a space so must be inside some kind of block.
                        // And everything between quotes is part of a block
                        inBlock = true
                    }

                    if ((state == UpdateBlockSequence.KEY || state == UpdateBlockSequence.VALUE) && inBlock && c != '"' && c != '\'') {
                        string[stringIndex++] = c
                        if(stringIndex == MAX_STRING_SIZE) error("extractBeforeAndAfterMap: string to long to fit in buffer ($sqlRedo)")
                    }
                }
            }
        }
        return Pair(beforeMap, afterMap)
    }
}

private enum class InsertBlockSequence {
    FIELDS,
    VALUES
}

private enum class UpdateBlockSequence {
    KEY,
    SEPERATOR_AFTER_KEY,
    VALUE,
    SEPARATOR_AFTER_VALUE
}


`

File: debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerQueryResultProcessor.java (hunk 9 above)
Comment (MilovdZee) [reply debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerQueryResultProcessor.java:181]: done

File: debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java [hunk 31]
```diff
@@ -125,6 +138,37 @@
     public static final Field SERVER_NAME = RelationalDatabaseConnectorConfig.SERVER_NAME
             .withValidation(CommonConnectorConfig::validateServerNameIsDifferentFromHistoryTopicName);
 
+    public static final Field CONNECTOR_ADAPTER = Field.create(DATABASE_CONFIG_PREFIX + "connection.adapter")
```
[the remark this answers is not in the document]
Naros: This one is already included.

debezium-connector-oracle/pom.xml
```diff
@@ -93,6 +104,11 @@
         </dependency>
     </dependencies>
 
+    <properties>
+        <adapter.name>xstream</adapter.name>
+        <version.oracle.driver>12.2.0.1</version.oracle.driver>
```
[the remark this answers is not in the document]
Naros: Ah yes, its defined in the incubator parent pom.xml so I'll remove this.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
```diff
@@ -96,6 +107,89 @@ public void resetSessionToCdb() {
...
+        }
+        return tableIds;
+    }
+
+    @Override // todo replace metadata with a query
+    protected Optional<ColumnEditor> readTableColumn(ResultSet columnMetadata, TableId tableId, ColumnNameFilter columnFilter) throws SQLException {
+        return super.readTableColumn(columnMetadata, tableId, columnFilter);
+    }
+
+    // todo replace metadata with something like this
```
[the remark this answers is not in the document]
Naros: Captured as [DBZ-2528](https://issues.redhat.com/browse/DBZ-2528)

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDatabaseSchema.java
Naros: It wouldn't be needed if we change `RelationalDatabaseSchema#tables()` as `public` rather than `protected`.  I wonder if it would simply make more sense to leave this for now as only Oracle needs access to the `Tables` object this way and we look at making changes to Oracle to avoid needing access to this to follow how other connectors behave instead?

Naros:
I am unable to reply to the comments on {{SqlUtils}} directly for some reason in the GitHub UI.  I've added [DBZ-2529](https://issues.redhat.com/browse/DBZ-2529) and [DBZ-2530](https://issues.redhat.com/browse/DBZ-2530).

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerMetrics.java
Naros: Done

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHelper.java
Naros: Done

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/jsqlparser/SimpleDmlParser.java
Naros: Looking at the code here I think honestly that `table` could be factored out of being a class member and instead simply be an argument that is passed to the methods that actually use it instead; however I'd rather do this as a code-cleanup task later because I'm not sure with the new parser implementation from Andrey & team, how much of this code would remain.

I've made the instance variable changes to those that can be final; and I logged a refactor task [DBZ-2532](https://issues.redhat.com/browse/DBZ-2532).

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/antlr/OracleDmlParser.java
Naros: Done

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/jsqlparser/SimpleDmlParser.java
Naros: I've included this as part of [DBZ-2532](https://issues.redhat.com/browse/DBZ-2532).  Please feel free to leave any additional comments there as I've referenced the above code snippet for reference to this discussion.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
Naros: Logged as [DBZ-2530](https;//issues.redhat.com/browse/DBZ-2530).

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
Naros: Logged as [DBZ-2531](https://issues.redhat.com/browse/DBZ-2531).

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
Naros: Logged as [DBZ-2529](https://issues.redhat.com/browse/DBZ-2529).

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorTask.java
Naros: Done

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/TransactionalBufferMetrics.java
Naros: Done

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerMetrics.java
Naros: I've marked these as final.  I'll update the DBZ-137 docs PR and add information about metrics there.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
Naros: I've removed the pertinent code here and I'll re-add it as part of DBZ-2525.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
Naros: Code removed & re-implement as suggested in the other issue.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
Naros: Logged as [DBZ-2533](https://issues.redhat.com/browse/DBZ-2533).

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSourceInfoStructMaker.java
Naros: Added in the DBZ-137 docs update PR.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
Naros: I reverted this change.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
Naros: Reverted this change.

Naros:
@gunnarmorling This should be ready for one final review.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDatabaseSchema.java
gunnarmorling: Sounds good.

gunnarmorling:
Rebased and applied!

Thank you so much to everyone involved: @AndreyIg, @jchipmunk, @mallikankati, @MilovdZee, @Naros, @jpechane, and others who I may have forgotten!

There's a number of follow-up tasks, but I think what we have is in a state now where people can test it out; follow-up tasks are logged in Jira (linked to DBZ-137). If there's other things, please don't hesitate to comment and/or open more issues. This will still take some time to bake, but the foundation is there. Thanks again!

PR #188, opened 2020-09-01
DBZ-2462 avoid divisive languange in incubator connectors

rk3rn3r:
* add include/exclude list properties for cassandra, db2 and oracle incubator connectors
* deprecate whitelist/blacklist properties
* clean some code + minor housekeeping

fixes https://issues.redhat.com/browse/DBZ-2462

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,16 +23,20 @@ Once connectors are deemed mature enough, they may be promoted into the Debezium
 
 Please see the [README.md](https://github.com/debezium/debezium#building-debezium) in the main repository for general instructions on building Debezium from source (prerequisites, usage of Docker etc).
 
-### Building the Cassandra connector
+### Building the Cassandra 3.x connector
 
-In order to build the Cassandra connector you'll need JDK <= 9 because Cassandra
-doesn't support Java versions above Java 9.
+In order to build the Cassandra connector you'll need JDK 8 because Cassandra 3.x
+doesn't support Java versions above Java 8. That also means dependencies like
+`debezium-core` have to be built as Java 8 bytecode version 52.0 as well,
+either by compiling it with Java 8 or specifying Java 8 bytecode generation
+on newer versions of Java.
 
 Then the Cassandra connector can be built like so:
 
     $ mvn clean install -am -pl debezium-connector-cassandra
     
-If you have multiple Java installation on your machine you can select the correct version by setting JAVA_HOME env var:
+If you have multiple Java installation on your machine you can select the correct
+version by setting JAVA_HOME env var:
 
     $ JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64 mvn clean install -am -pl debezium-connector-cassandra
 
--- a/README.md
+++ b/README.md
@@ -23,6 +23,19 @@ Once connectors are deemed mature enough, they may be promoted into the Debezium
 
 Please see the [README.md](https://github.com/debezium/debezium#building-debezium) in the main repository for general instructions on building Debezium from source (prerequisites, usage of Docker etc).
 
+### Building the Cassandra connector
+
+In order to build the Cassandra connector you'll need JDK <= 9 because Cassandra
+doesn't support Java versions above Java 9.
+
+Then the Cassandra connector can be built like so:
+
+    $ mvn clean install -am -pl debezium-connector-cassandra
+    
+If you have multiple Java installation on your machine you can select the correct version by setting JAVA_HOME env var:
+
+    $ JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64 mvn clean install -am -pl debezium-connector-cassandra
+
 ### Building and testing the Db2 connector
 
 Running `mvn install` will compile all code and run the unit and integration tests. If there are any compile problems or any of the unit tests fail, the build will stop immediately. Otherwise, the command will continue to create the module's artifacts, create the Docker image with DB2 and custom scripts, start the Docker container, run the integration tests, stop the container (even if there are integration test failures), and run checkstyle on the code. If there are still no problems, the build will then install the module's artifacts into the local Maven repository.
@@ -69,10 +82,10 @@ making these parameters potentially obsolete.
 ```
 
 By default, Debezium will ignore some admin tables in Oracle 12c.
-But as those tables are different in Oracle 11g, Debezium will report an error on those tables. So when use Debezium on Oracle 11g, use table white list (remember to use lower case):
+But as those tables are different in Oracle 11g, Debezium will report an error on those tables. So when use Debezium on Oracle 11g, use `table.include.list` (remember to use lower case):
 
 ```json
-"table.whitelist":"orcl\\.debezium\\.(.*)"
+"table.include.list": "orcl\\.debezium\\.(.*)"
 ```
 
 Making this configuration obsolete is tracked under [DBZ-1045](https://issues.jboss.org/browse/DBZ-1045).
--- a/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
+++ b/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
@@ -13,7 +13,9 @@ import java.util.Properties;
 import java.util.stream.Collectors;
 
 import org.apache.kafka.clients.producer.ProducerConfig;
+import org.apache.kafka.common.config.ConfigDef.Importance;
 import org.apache.kafka.common.config.ConfigDef.Type;
+import org.apache.kafka.common.config.ConfigDef.Width;
 import org.apache.kafka.common.serialization.ByteArraySerializer;
 import org.apache.kafka.connect.storage.Converter;
 
@@ -241,7 +243,25 @@ public class CassandraConnectorConfig extends CommonConnectorConfig {
      * A comma-separated list of fully-qualified names of fields that should be excluded from change event message values.
      * Fully-qualified names for fields are in the form {@code <keyspace_name>.<field_name>.<nested_field_name>}.
      */
-    public static final Field FIELD_BLACKLIST = Field.create("field.blacklist").withType(Type.STRING);
+    public static final Field FIELD_EXCLUDE_LIST = Field.create("field.exclude.list")
+            .withDisplayName("Exclude Fields")
+            .withType(Type.STRING)
+            .withWidth(Width.LONG)
+            .withImportance(Importance.MEDIUM)
+            .withInvisibleRecommender()
+            .withDescription("Regular expressions matching columns to include in change events");
+
+    /**
+     * Old, backwards-compatible "blacklist" property.
+     */
+    @Deprecated
+    public static final Field FIELD_BLACKLIST = Field.create("field.blacklist")
+            .withDisplayName("Deprecated: Exclude Fields")
+            .withType(Type.STRING)
+            .withWidth(Width.LONG)
+            .withImportance(Importance.LOW)
+            .withInvisibleRecommender()
+            .withDescription("Regular expressions matching columns to include in change events (deprecated, use \"" + FIELD_EXCLUDE_LIST.name() + "\" instead)");
 
     /**
      * Instead of parsing commit logs from CDC directory, this will look for the commit log with the
@@ -413,12 +433,12 @@ public class CassandraConnectorConfig extends CommonConnectorConfig {
         return Duration.ofMillis(ms);
     }
 
-    public String[] fieldBlacklist() {
-        String hosts = this.getConfig().getString(FIELD_BLACKLIST);
-        if (hosts == null) {
+    public String[] fieldExcludeList() {
+        String fieldExcludeList = this.getConfig().getFallbackStringProperty(FIELD_EXCLUDE_LIST, FIELD_BLACKLIST);
+        if (fieldExcludeList == null) {
             return new String[0];
         }
-        return hosts.split(",");
+        return fieldExcludeList.split(",");
     }
 
     /**
--- a/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
+++ b/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
@@ -7,6 +7,8 @@ package io.debezium.connector.cassandra;
 
 import java.time.Duration;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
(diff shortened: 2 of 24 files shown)


debezium-connector-db2/src/main/java/io/debezium/connector/db2/Db2ConnectorConfig.java
```diff
@@ -279,14 +279,17 @@ public static SnapshotIsolationMode parse(String value, String defaultValue) {
...
+            Db2ConnectorConfig.DATABASE_HISTORY,
+            Db2ConnectorConfig.TABLE_WHITELIST,
+            Db2ConnectorConfig.TABLE_INCLUDE_LIST,
+            Db2ConnectorConfig.TABLE_BLACKLIST,
+            Db2ConnectorConfig.TABLE_EXCLUDE_LIST,
+            Db2ConnectorConfig.TABLE_IGNORE_BUILTIN,
+            Db2ConnectorConfig.COLUMN_BLACKLIST,
+            Db2ConnectorConfig.COLUMN_EXCLUDE_LIST,
+            Db2ConnectorConfig.DECIMAL_HANDLING_MODE,
+            Db2ConnectorConfig.TIME_PRECISION_MODE,
```
rk3rn3r, line 292: Beside adding the new include/exclude list properties I did some housekeeping here to move the parameters to the ConnectorConfig class related to that connector in order to local rewrites work and from architectural perspective we have less dependencies to "the outside".

debezium-connector-db2/src/main/java/io/debezium/connector/db2/Db2ConnectorConfig.java
```diff
@@ -302,18 +305,21 @@ public static ConfigDef configDef() {
...
-                RelationalDatabaseConnectorConfig.TABLE_IGNORE_BUILTIN,
+                KafkaDatabaseHistory.RECOVERY_POLL_INTERVAL_MS, Db2ConnectorConfig.DATABASE_HISTORY);
+        Field.group(config, "Events", Db2ConnectorConfig.TABLE_WHITELIST,
+                Db2ConnectorConfig.TABLE_INCLUDE_LIST,
+                Db2ConnectorConfig.TABLE_BLACKLIST,
+                Db2ConnectorConfig.TABLE_EXCLUDE_LIST,
+                Db2ConnectorConfig.COLUMN_BLACKLIST,
+                Db2ConnectorConfig.COLUMN_EXCLUDE_LIST,
+                Db2ConnectorConfig.SNAPSHOT_SELECT_STATEMENT_OVERRIDES_BY_TABLE,
+                Db2ConnectorConfig.TABLE_IGNORE_BUILTIN,
```
rk3rn3r, line 316: Beside adding the new include/exclude list properties I did some housekeeping here to move the parameters to the ConnectorConfig class related to that connector in order to local rewrites work and from architectural perspective we have less dependencies to "the outside".

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -138,11 +138,13 @@
...
-            RelationalDatabaseConnectorConfig.TABLE_BLACKLIST,
-            RelationalDatabaseConnectorConfig.TABLE_IGNORE_BUILTIN,
-            RelationalDatabaseConnectorConfig.MSG_KEY_COLUMNS,
+            OracleConnectorConfig.DATABASE_HISTORY,
+            OracleConnectorConfig.TABLE_WHITELIST,
+            OracleConnectorConfig.TABLE_INCLUDE_LIST,
+            OracleConnectorConfig.TABLE_BLACKLIST,
+            OracleConnectorConfig.TABLE_EXCLUDE_LIST,
+            OracleConnectorConfig.TABLE_IGNORE_BUILTIN,
+            OracleConnectorConfig.MSG_KEY_COLUMNS,
```
rk3rn3r, line 147: Beside adding the new include/exclude list properties I did some housekeeping here to move the parameters to the ConnectorConfig class related to that connector in order to local rewrites work and from architectural perspective we have less dependencies to "the outside".

(also in several other cases)

README.md
```diff
@@ -23,6 +23,19 @@ Once connectors are deemed mature enough, they may be promoted into the Debezium
 
 Please see the [README.md](https://github.com/debezium/debezium#building-debezium) in the main repository for general instructions on building Debezium from source (prerequisites, usage of Docker etc).
 
+### Building the Cassandra connector
+
+In order to build the Cassandra connector you'll need JDK <= 9 because Cassandra
```
gunnarmorling, line 28: Wow, is this true indeed? Do you have a reference?

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
```diff
@@ -241,7 +243,25 @@
...
      * Fully-qualified names for fields are in the form {@code <keyspace_name>.<field_name>.<nested_field_name>}.
      */
-    public static final Field FIELD_BLACKLIST = Field.create("field.blacklist").withType(Type.STRING);
+    public static final Field FIELD_EXCLUDE_LIST = Field.create("field.exclude.list")
+            .withDisplayName("Exclude Fields")
+            .withType(Type.STRING)
+            .withWidth(Width.LONG)
+            .withImportance(Importance.MEDIUM)
+            .withInvisibleRecommender()
+            .withDescription("Regular expressions matching columns to include in change events");
```
gunnarmorling, line 252: s/columns/fields/?

gunnarmorling: Same below.

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
```diff
@@ -413,12 +433,12 @@ public Duration snapshotPollIntervalMs() {
         return Duration.ofMillis(ms);
     }
 
-    public String[] fieldBlacklist() {
-        String hosts = this.getConfig().getString(FIELD_BLACKLIST);
-        if (hosts == null) {
+    public String[] fieldExcludeList() {
```
gunnarmorling, line 436: Let's use List.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
gunnarmorling: WDYM by "moving" exactly? Just referencing those fileds via `OracleConnectorConfig` instead of say `HistorizedRelationalDatabaseConnectorConfig`? If so, I don't think it's a strong improvement; in fact I'd even expect a warning as static members should always be accessed via the defining type instead of via sub-class. I'd prefer to revert this change.

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/OracleConnectorFilterIT.java
```diff
@@ -69,41 +68,41 @@ public void before() throws SQLException {
                 "GRANT CREATE TABLE TO debezium2",
                 "GRANT CREATE SEQUENCE TO debezium2",
                 "ALTER USER debezium2 QUOTA 100M ON users",
-                "create table debezium2.table2 (id numeric(9,0) not null,name varchar2(1000),primary key (id))",
-                "create table debezium2.nopk (id numeric(9,0) not null)",
-                "GRANT ALL PRIVILEGES ON debezium2.table2 to debezium",
-                "GRANT SELECT ON debezium2.table2 to " + TestHelper.CONNECTOR_USER,
-                "GRANT SELECT ON debezium2.nopk to " + TestHelper.CONNECTOR_USER,
+                "CREATE TABLE debezium2.table2 (id NUMERIC(9,0) NOT NULL, name VARCHAR2(1000), PRIMARY KEY (id))",
```
gunnarmorling, line 71: Very thorough :)

gunnarmorling (commented):
LGTM overall. A few minor comments/suggestions inline.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
rk3rn3r: I am not such a Java expert like you are because I didn't do professional Java for so many years like you, but I think what you mean is that you don't access a static by an instance, like:
```
class A {
    static int x = 2;
    ...
}

A a = new A();
a.x;
```
Do you have a link that explains it, because neither IDE nor compilation showed an error/warning so I guessed it's valid?

I think it's valid, especially when you mean to override parent properties in `OracleConnectorConfig` you expect the local values to be used. I thought this is more like a super/this replacement in a static scope?

README.md
rk3rn3r: LOL it didn't compile for me with JDK 11 and I had to download and use JDK 8 for building it. I think that source was a stackoverflow response.
Now that you were asking I found https://cassandra.apache.org/doc/latest/new/java11.html#support-matrix and that there's `ant -Duse.jdk=11`. I will try if that helps and document that instead.

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
rk3rn3r: that's not my code. will fix it though

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
rk3rn3r: more housekeeping! I like!

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
gunnarmorling: You are right, it doesn't generate an error. Note you cannot override a static member though, rather it will be _hidden_ when declaring an identical member in a sub-class.

That said, I preferred the original way of referencing those props via the defining (super-)class. If nothing else, it keeps the code more generic, e.g. making it easier to extract and reuse (or just copy and paste), if needed.

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/OracleConnectorFilterIT.java
rk3rn3r: When the tests didn't work for me I was trying to understand what's going on and thought it would be my code changes. To clearer understand what's going on I did that.
Finally it turned out I just didn't re-setup the docker container often enough inside the vagrant VM LOL.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
rk3rn3r: But using `RelationalDatabaseConnectorConfig.SERVER_NAME` will always returns the definition from `RelationalDatabaseConnectorConfig`, correct?
And when I use `RelationalDatabaseConnectorConfig.SERVER_NAME` everywhere inside the Oracle connector module I have a dependency everywhere to that external resource. Wouldn't it be much easier to have only the dependency to `OracleConnectorConfig` which resided in the module and handle everything there?
Futhermore I believe it makes it more predictable when I override (also it technically only hides the original value) something in `OracleConnectorConfig` that it is automatically used everywhere and I don't have to switch it on all occasions manually? Gives the connector config a better meaning too?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -291,7 +295,7 @@ public static OracleVersion parse(String value, String defaultValue) {
     /**
      * The set of predefined SnapshotMode options or aliases.
      */
-    public static enum SnapshotMode implements EnumeratedValue {
+    public enum SnapshotMode implements EnumeratedValue {
```
jpechane, line 298: @rk3rn3r Why is the `static` dropped here?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
gunnarmorling: > Wouldn't it be much easier to have only the dependency to OracleConnectorConfig

But that's the thing, the nature of the dependency doesn't change a bit. The code still - rightly so - depends on the property definition from the super-class, it only gets swepped under the rug, if you will. I strongly prefer to rever this change, in any case it's unrelated to the matter at hand. Thanks!

README.md
rk3rn3r: https://issues.apache.org/jira/browse/CASSANDRA-9608 introduces Java 11 compatibility with Cassandra 4.0. The connector currently uses Cassandra 3.x

debezium-connector-db2/src/main/java/io/debezium/connector/db2/Db2ConnectorConfig.java
rk3rn3r: Will be reverted

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
rk3rn3r: Will be reverted

debezium-connector-db2/src/main/java/io/debezium/connector/db2/Db2ConnectorConfig.java
rk3rn3r: Done

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
rk3rn3r: Done

README.md
rk3rn3r: Done

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
rk3rn3r: Done

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CassandraConnectorConfig.java
rk3rn3r: Done

gunnarmorling:
LGTM, applied. Thanks, a lot, @jpechane!

@bingqinzhou, quick heads-up for you: the whitelist/blacklist option has been renamed for Cassandra (as for all other Debezium connectors) for the sake of a more inclusive language. The old names can be used, but a warning will be raised. So you might want to update your config of running instances at some point.

PR #190, opened 2020-09-04
[DBZ-2498]: Fix error with frozen collection type column

bingqinzhou:
https://issues.redhat.com/browse/DBZ-2498


gunnarmorling:
Squashed and applied. Thx, @bingqinzhou!

PR #191, opened 2020-09-09
[DBZ-2510]: Add more logs

bingqinzhou:
Log keyspace, table and column when an error occurs.

### DIFF
--- a/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CommitLogReadHandlerImpl.java
+++ b/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CommitLogReadHandlerImpl.java
@@ -425,7 +425,7 @@ public class CommitLogReadHandlerImpl implements CommitLogReadHandler {
                     Object value;
                     Object deletionTs = null;
                     AbstractType abstractType = cd.type;
-                    if (abstractType.isCollection()) {
+                    if (abstractType.isCollection() && abstractType.isMultiCell()) {
                         ComplexColumnData ccd = row.getComplexColumnData(cd);
                         value = CassandraTypeDeserializer.deserialize((CollectionType) abstractType, ccd);
                     }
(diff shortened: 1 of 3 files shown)


debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/CommitLogReadHandlerImpl.java
```diff
@@ -395,23 +395,30 @@ private void handleRowModifications(Row row, RowType rowType, PartitionUpdate pu
...
-            CellData cellData = new CellData(name, value, null, CellData.ColumnType.PARTITION);
-            after.addCell(cellData);
+            try {
+                String name = cd.name.toString();
+                Object value = partitionKeys.get(cd.position());
+                CellData cellData = new CellData(name, value, null, CellData.ColumnType.PARTITION);
+                after.addCell(cellData);
+            }
+            catch (Exception e) {
+                LOGGER.error("Failed to populate Column {} with Type {} of Table {} in KeySpace {}.",
```
gunnarmorling, line 405: Let's better either log _or_ rethrow, but not both at the same time, as it'll clutter log files. So I'd suggest to wrap the original exception in a `DebeziumException`, adding the message with the additional context you'd like to propagate. It'll all be logged at once then.

gunnarmorling (changes requested):
Thanks, @bingqinzhou! One comment inline which applies to all the changed code parts essentially. Thanks!

gunnarmorling:
@bingqinzhou, LGTM now. Could you only squash everything into one commit and rebase it to the current master (in general, let's work with rebases over merges for the sake of a linear and thus cleaner history)? Thanks!

gunnarmorling:
@bingqinzhou, one more question actually: why is DBZ-2498 reverted here?

bingqinzhou:
> @bingqinzhou, one more question actually: why is DBZ-2498 reverted here?

Not sure, it seems that 462a5f4, 869b513, 8efe8bc, 7428150 are actually commits from #190 instead of this one. The commit history in my local master seems be messed up, I'm deleting this PR and created #193 with a cleaner commit history.

gunnarmorling:
Gotcha, no worries. Merged #193 and will close this one. Thanks, @bingqinzhou!

PR #194, opened 2020-09-16
[DBZ-2512]: Fix error with collection type of UUID

bingqinzhou:
This PR is to fix the issue in https://issues.redhat.com/browse/DBZ-2512:

After deserializing UUID column from Cassandra, we need convert the deserialized UUID value to String in order to fit in Kafka Connect Schema which doesn't have a UUID type. We need to add the same step to convert each UUID element in deserialized Collection<UUID> column to String as well, otherwise we'll get error when sending out change event to kafka for Collection<UUID> columns.

### DIFF
--- a/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/transforms/type/deserializer/LogicalTypeDeserializer.java
+++ b/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/transforms/type/deserializer/LogicalTypeDeserializer.java
@@ -1,0 +1,12 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.cassandra.transforms.type.deserializer;
+
+import org.apache.cassandra.db.marshal.AbstractType;
+
+public abstract class LogicalTypeDeserializer extends TypeDeserializer {
+    public abstract Object convertDeserializedValue(AbstractType<?> abstractType, Object value);
+}
--- a/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/transforms/type/deserializer/LogicalTypeDeserializer.java
+++ b/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/transforms/type/deserializer/LogicalTypeDeserializer.java
@@ -7,6 +7,9 @@ package io.debezium.connector.cassandra.transforms.type.deserializer;
 
 import org.apache.cassandra.db.marshal.AbstractType;
 
+/**
+ * For deserializing logical-type columns in Cassandra, like UUID, TIMEUUID, Duration, etc.
+ */
 public abstract class LogicalTypeDeserializer extends TypeDeserializer {
 
     /**
@@ -17,12 +20,4 @@ public abstract class LogicalTypeDeserializer extends TypeDeserializer {
      */
     public abstract Object convertDeserializedValue(AbstractType<?> abstractType, Object value);
 
-    /**
-     * Check if a TypeDeserializer extends from LogicalTypeDeserializer
-     * @param typeDeserializer TypeDeserializer
-     * @return true if the TypeDeserializer extends from LogicalTypeDeserializer
-     */
-    public static boolean isParentOf(TypeDeserializer typeDeserializer) {
-        return LogicalTypeDeserializer.class.isAssignableFrom(typeDeserializer.getClass());
-    }
 }
--- a/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/transforms/type/deserializer/LogicalTypeDeserializer.java
+++ b/debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/transforms/type/deserializer/LogicalTypeDeserializer.java
@@ -8,5 +8,21 @@ package io.debezium.connector.cassandra.transforms.type.deserializer;
 import org.apache.cassandra.db.marshal.AbstractType;
 
 public abstract class LogicalTypeDeserializer extends TypeDeserializer {
+
+    /**
+     * Convert the deserialized value from Cassandra to an object that fits kafka schema type
+     * @param abstractType the {@link AbstractType} of a column in cassandra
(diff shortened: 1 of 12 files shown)


bingqinzhou:
@gunnarmorling May I get a review for this PR please? It is for fixing the error we encountered when sending out change events to kafka with column type of Collection<UUID>. Please let me know if you need more context or additional changes from me, thank you!

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/transforms/type/deserializer/LogicalTypeDeserializer.java
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.cassandra.transforms.type.deserializer;
+
+import org.apache.cassandra.db.marshal.AbstractType;
+
+public abstract class LogicalTypeDeserializer extends TypeDeserializer {
```
gunnarmorling, line 10: Could you add a class comment?

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/transforms/type/deserializer/LogicalTypeDeserializer.java
```diff
@@ -0,0 +1,28 @@
...
+     */
+    public abstract Object convertDeserializedValue(AbstractType<?> abstractType, Object value);
+
+    /**
+     * Check if a TypeDeserializer extends from LogicalTypeDeserializer
+     * @param typeDeserializer TypeDeserializer
+     * @return true if the TypeDeserializer extends from LogicalTypeDeserializer
+     */
+    public static boolean isParentOf(TypeDeserializer typeDeserializer) {
+        return LogicalTypeDeserializer.class.isAssignableFrom(typeDeserializer.getClass());
```
gunnarmorling, line 26: Why not simply call `typeDeserializer instanceof LogicalTypeDeserializer`? Is this method needed even?

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/transforms/type/deserializer/SetTypeDeserializer.java
```diff
@@ -25,29 +25,47 @@
...
+        return Values.convertToList(getSchemaBuilder(setType).build(), deserializedList);
+    }
+
+    /**
+     * If elements in a deserialized set is LogicalType, convert each element to fit in Kafka Schema type
+     * @param abstractType the {@link AbstractType} of a column in Cassandra
+     * @param deserializedSet Set deserialized from Cassandra
+     * @return A deserialized list from Cassandra with each element that fits in Kafka Schema type
+     */
+    private List<Object> convertDeserializedElementsIfNecessary(AbstractType<?> abstractType, Set<?> deserializedSet) {
```
gunnarmorling, line 58: How comes a set goes in but a list goes out? Would it make sense to emit sets?

gunnarmorling (commented):
@bingqinzhou not too much to say on the change, apart from a few small suggestions. Only I was expecting a changed/new test, too? Something related to tue UUID problem reported in the original Jira issue?

gunnarmorling:
Ah, and the last commit made the log limit on CI to overflow. Can you modify log4j.properties to reduce verbosity a bit? Thanks, @bingqinzhou!

gunnarmorling:
Hey @bingqinzhou, quick heads-up: we're planning to do 1.3.0.CR1 tomorrow, so ideally this would be wrapped up by then. If not, I think we still can add it to 1.3.0.Final (due next week), risk should be low in the large scheme of things.

bingqinzhou:
> Hey @bingqinzhou, quick heads-up: we're planning to do 1.3.0.CR1 tomorrow, so ideally this would be wrapped up by then. If not, I think we still can add it to 1.3.0.Final (due next week), risk should be low in the large scheme of things.

I'll wrap it up by 1.3.0.Final next week, thank you!

gunnarmorling:
Ok, thx for the heads-up, @bingqinzhou. We're aiming at Wednesday (Sept. 30) for that.

gunnarmorling:
LGTM now, merged. Thanks, @bingqinzhou!

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/transforms/type/deserializer/LogicalTypeDeserializer.java
bingqinzhou: > Could you add a class comment?

Added!

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/transforms/type/deserializer/LogicalTypeDeserializer.java
bingqinzhou: > Why not simply call `typeDeserializer instanceof LogicalTypeDeserializer`? Is this method needed even?

Good point! Used instanceof instead

debezium-connector-cassandra/src/main/java/io/debezium/connector/cassandra/transforms/type/deserializer/SetTypeDeserializer.java
bingqinzhou: > How comes a set goes in but a list goes out? Would it make sense to emit sets?

We don't have a convertToSet method in Kafka Connect library, there's only convertToList, so we have to convert a set to a list in kafka here.

bingqinzhou:
> @gunnarmorling May I get a review for this PR please? It is for fixing a critical error we encountered when sending out change events to kafka with column type of Collection[UUID]. Please let me know if you need more context or additional changes from me, thank you!


> @bingqinzhou not too much to say on the change, apart from a few small suggestions. Only I was expecting a changed/new test, too? Something related to tue UUID problem reported in the original Jira issue?

Good point! Added unit test for Collection[UUID] type deserialization.

PR #195, opened 2020-09-17
DBZ-2555 Fix LogMiner schema change event dispatch

Naros:
https://issues.redhat.com/browse/DBZ-2555

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/RowMapper.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/RowMapper.java
@@ -169,7 +169,7 @@ public class RowMapper {
     }
 
     public static TableId getTableId(String catalogName, ResultSet rs) throws SQLException {
-        return new TableId(catalogName.toUpperCase(), rs.getString(SEG_OWNER).toUpperCase(), rs.getString(TABLE_NAME).toUpperCase());
+        return new TableId(catalogName, rs.getString(SEG_OWNER), rs.getString(TABLE_NAME));
     }
 
 }
--- a/debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/logminer/RowMapperTest.java
+++ b/debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/logminer/RowMapperTest.java
@@ -26,6 +26,7 @@ import org.mockito.Mockito;
 import io.debezium.connector.oracle.junit.SkipTestDependingOnAdapterNameRule;
 import io.debezium.connector.oracle.junit.SkipWhenAdapterNameIsNot;
 import io.debezium.connector.oracle.junit.SkipWhenAdapterNameIsNot.AdapterName;
+import io.debezium.doc.FixFor;
 import io.debezium.relational.TableId;
 
 @SkipWhenAdapterNameIsNot(value = AdapterName.LOGMINER)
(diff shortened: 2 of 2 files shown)


debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/logminer/RowMapperTest.java
```diff
@@ -173,4 +174,24 @@ public void testGetTableId() throws SQLException {
...
+        Mockito.when(rs.getString(7)).thenReturn("table");
+        TableId tableId = RowMapper.getTableId("CATALOG", rs);
+        assertThat(tableId.toString().equals("CATALOG.Schema.table")).isTrue();
+        verify(rs).getString(8);
+        Mockito.when(rs.getString(8)).thenThrow(SQLException.class);
+
+        tableId = null;
+        try {
+            tableId = RowMapper.getTableId("catalog", rs);
+            assertThat(1 == 2).isTrue();
```
rk3rn3r, line 191: :thinking:  assertTrue(1 == 2) ? you want to make that fail, right?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/RowMapper.java
```diff
@@ -169,7 +169,7 @@ private static void logError(TransactionalBufferMetrics metrics, SQLException e,
     }
 
     public static TableId getTableId(String catalogName, ResultSet rs) throws SQLException {
-        return new TableId(catalogName.toUpperCase(), rs.getString(SEG_OWNER).toUpperCase(), rs.getString(TABLE_NAME).toUpperCase());
+        return new TableId(catalogName, rs.getString(SEG_OWNER), rs.getString(TABLE_NAME));
```
gunnarmorling, line 172: That seems quite disruptive, don't we expose all names in upper-case for Oracle?

gunnarmorling (commented):
One question inline.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/RowMapper.java
Naros: Unless I missed something looking over the code, I don't think we explicitly control the case in the xstreams impl but rather we simply emit the events based on the case of the identifiers returned by Oracle itself.

Oracle identifiers are by default stored in upper-case so long as the identifier isn't quoted, see [here](https://docs.oracle.com/cd/B19306_01/server.102/b14200/sql_elements008.htm).  But for identifiers that contain certain characters, Oracle does require that those names be double-quoted.  We added support for things such as hyphens as a part of DBZ-XXXX and the test, found [here](https://github.com/debezium/debezium-incubator/blob/master/debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/OracleConnectorIT.java#L144), is what triggered me actually making this case change for LogMiner.

In short, the table is created as `"my-table"` and so Oracle mandates we always refer to this table with that explicit case, it will also always return that object name with lower-case since that's how it's explicitly created.  Additionally, Oracle also supports creating differing objects with the same named identifier so long as you explicitly quote the objects to control the case, thus allowing multiple tables with `"employees"`, `"Employees"`, and `"EMPLOYEES"` as an example.

So I actually believe what LogMiner is doing here is wrong and we should respect the explicit case Oracle gives the table metadata as and emit events accordingly.  FWIW, the referenced test case above works fine with XStreams and the emitted event actually uses a lower-case table-name due to the explicit quotations.

What do you think?

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/logminer/RowMapperTest.java
Naros: Pushed a commit to fix this.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/RowMapper.java
gunnarmorling: Yes, this makes sense. I had missed originally that this change doesn't impact the existing XStreams-based implementation. Will merge.

PR #196, opened 2020-09-17
DBZ-2531 Implement connection retry for Oracle

Naros:
https://issues.redhat.com/browse/DBZ-2531

NOTE:
While this PR is independent of PR #195, I found that LogMiner has the potential to fail to lookup a table's schema due to incorrect handling of the table-name case.  Therefore its possible when testing this PR against LogMiner specifically, a single test will fail.  By incorporating the changes from PR #195, both connection adapters pass.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleErrorHandler.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleErrorHandler.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.oracle;
+
+import java.io.IOException;
+import java.sql.SQLRecoverableException;
+
+import io.debezium.connector.base.ChangeEventQueue;
+import io.debezium.pipeline.ErrorHandler;
+
+import oracle.net.ns.NetException;
+
+/**
+ * Error handle for Oracle.
+ *
+ * @author Chris Cranford
+ */
+public class OracleErrorHandler extends ErrorHandler {
+
+    public OracleErrorHandler(String logicalName, ChangeEventQueue<?> queue) {
+        super(OracleConnector.class, logicalName, queue);
+    }
+
+    @Override
+    protected boolean isRetriable(Throwable throwable) {
+        if (throwable.getMessage() == null || throwable.getCause() == null) {
+            return false;
+        }
+
+        return throwable.getMessage().startsWith("ORA-03135") || // connection lost
+                throwable.getMessage().startsWith("ORA-12543") || // TNS:destination host unreachable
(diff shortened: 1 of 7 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleErrorHandler.java
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        return throwable.getMessage().startsWith("ORA-03135") || // connection lost
+                throwable.getMessage().startsWith("ORA-12543") || // TNS:destination host unreachable
+                throwable.getMessage().startsWith("ORA-00604") || // error occurred at recursive SQL level 1
+                throwable.getMessage().startsWith("ORA-01089") || // Oracle immediate shutdown in progress
+                throwable.getCause() instanceof IOException ||
+                throwable instanceof SQLRecoverableException ||
+                throwable.getMessage().toUpperCase().startsWith("NO MORE DATA TO READ FROM SOCKET") ||
+                throwable.getCause().getCause() instanceof NetException;
```
gunnarmorling, line 40: Might be prone to NPE if there's not more than one cause in the chain?

gunnarmorling (commented):
Nice, LGTM. Just one small question inline.

gunnarmorling:
@Naros, pushed an unrelated commit to this PR as it wasn't worth logging a dedicated issue. I hope you don't mind the piggy-backing. Also, if you've adressed the comment above, please feel free to merge this one yourself. Thanks!

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleErrorHandler.java
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        return throwable.getMessage().startsWith("ORA-03135") || // connection lost
+                throwable.getMessage().startsWith("ORA-12543") || // TNS:destination host unreachable
+                throwable.getMessage().startsWith("ORA-00604") || // error occurred at recursive SQL level 1
+                throwable.getMessage().startsWith("ORA-01089") || // Oracle immediate shutdown in progress
+                throwable.getCause() instanceof IOException ||
+                throwable instanceof SQLRecoverableException ||
+                throwable.getMessage().toUpperCase().startsWith("NO MORE DATA TO READ FROM SOCKET") ||
+                (throwable.getCause().getCause() != null && throwable.getCause().getCause() instanceof NetException);
```
gunnarmorling, line 40:
```suggestion
                (throwable.getCause() != null && throwable.getCause().getCause() instanceof NetException);
```

@Naros there'd be an NPE if the direct cause of the `throwable` is null. No need to check the cause's cause, as `null instanceof ...` simply returns false.

gunnarmorling (approved):
LGTM, only log verbosity needs adjustment to prevent log overflow on CI.

gunnarmorling:
Thx, @Naros + @jpechane!

PR #197, opened 2020-09-25
DBZ-2591 take last processed lsn from topic rather than from db regis…

denisprog:
In case connector was not working for some time  - new changes were completed and registered in register table.
Connector starts and gets max lsn from register table. In this case changes were made during connector was down were skipped and only last(max) change was processed.
FIX: get max processed lsn from topic without comparing it with value in register table,


lga-zurich (approved):
As discussed, thanks for this.

PR #198, opened 2020-09-29
DBZ-2607 Fix NullPointerException when getting all table ids

Naros:
https://issues.redhat.com/browse/DBZ-2607

A regression was introduced by the LogMiner implementation that required `database.schema` to also be used for XStream.  This PR fixes this regression and does not require this configuration option to be provided when using XStreams.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
@@ -108,7 +108,7 @@ public class OracleConnection extends JdbcConnection {
     }
 
     protected Set<TableId> getAllTableIds(String catalogName, String schemaNamePattern, boolean isView) throws SQLException {
-
+        schemaNamePattern = schemaNamePattern == null ? "" : schemaNamePattern;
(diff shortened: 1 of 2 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
```diff
@@ -108,7 +108,7 @@ public void resetSessionToCdb() {
     }
 
     protected Set<TableId> getAllTableIds(String catalogName, String schemaNamePattern, boolean isView) throws SQLException {
-
+        schemaNamePattern = schemaNamePattern == null ? "" : schemaNamePattern;
```
gunnarmorling, line 111: Wouldn't it make more sense to omit the `owner` part altogether if schema pattern is null? I guess the query optimizer will discard the %% clause anyways, but seems a bit cleaner in terms of our code. WDYT?

gunnarmorling: Pushed one commit for that, and also for using a statement parameter (which we should have done so already before). I can't test though locally right now, so could you review and verify? Thx!

Naros: The change LGTM, thanks!

PR #203, opened 2020-10-14
[DBZ-2661] Force slf4j to use log4j instead of logback.

bingqinzhou:


jpechane:
@bingqinzhou Applied, thanks

PR #202, opened 2020-10-08
DBZ-2629 Fix LogMiner stopping due to ORA-01013

Naros:
https://issues.redhat.com/browse/DBZ-2629

This PR makes a small re-arrangement of SQL commands to avoid the `ALTER SESSION` resulting in an ORA-01013.

PR #204, opened 2020-10-19
DBZ-2624 Fix SQLExceptions being thrown while using LogMiner 

Naros:
https://issues.redhat.com/browse/DBZ-2624
Note: This PR is based on PR #202 and therefore should be merged afterward.


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectionFactory.java
```diff
@@ -5,32 +5,34 @@
...
-        String database = config.getDatabase();
-        String user = config.getUser();
-        String password = config.getPassword();
-
-        final String url = "jdbc:oracle:" + getDriverType(config) + ":@" + hostName + ":" + port + "/" + database;
-
-        return DriverManager.getConnection(url, user, password);
+        final String driverType = getDriverType(config);
+        if (DRIVER_TYPE_THIN.equals(driverType)) {
+            System.out.println("Configs: " + config.keys());
```
jpechane, line 26: @Naros Hi, I think this code was not run through formatter + checkstyle.

Naros: LOL, thanks.  I'll fix & rebase with master since you merged the other PR.

Naros: Fixed

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectionFactory.java
```diff
@@ -5,32 +5,33 @@
...
-
-        return DriverManager.getConnection(url, user, password);
+        final String driverType = getDriverType(config);
+        if (DRIVER_TYPE_THIN.equals(driverType)) {
+            if (!config.keys().contains(USE_FETCH_SIZE_WITH_LONG_COLUMN.toLowerCase())) {
+                config = JdbcConfiguration.adapt(config.edit().with(USE_FETCH_SIZE_WITH_LONG_COLUMN, "true").build());
+            }
+            return JdbcConnection.patternBasedFactory("jdbc:oracle:thin:@${hostname}:${port}/${dbname}").connect(config);
+        }
+        return JdbcConnection.patternBasedFactory("jdbc:oracle:oci:@{hostname}:${port}?${dbname}").connect(config);
```
jpechane, line 31: Why is here question mark as database name separater in URL and not slash?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectionFactory.java
```diff
@@ -5,32 +5,33 @@
...
-        int port = config.getPort();
-        String database = config.getDatabase();
-        String user = config.getUser();
-        String password = config.getPassword();
-
-        final String url = "jdbc:oracle:" + getDriverType(config) + ":@" + hostName + ":" + port + "/" + database;
-
-        return DriverManager.getConnection(url, user, password);
+        final String driverType = getDriverType(config);
+        if (DRIVER_TYPE_THIN.equals(driverType)) {
```
jpechane, line 25: I am a little bit confused here - `USE_FETCH_SIZE_WITH_LONG_COLUMN` is set to `true` by default. Why does it need to be overwritten here?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectionFactory.java
```diff
@@ -5,32 +5,33 @@
...
 import java.sql.Connection;
-import java.sql.DriverManager;
 import java.sql.SQLException;
 
 import io.debezium.connector.oracle.OracleConnectorConfig.ConnectorAdapter;
 import io.debezium.jdbc.JdbcConfiguration;
+import io.debezium.jdbc.JdbcConnection;
 import io.debezium.jdbc.JdbcConnection.ConnectionFactory;
 
 public class OracleConnectionFactory implements ConnectionFactory {
```
jpechane, line 16: BTW, as I think about that, do we need the indirect construction via `ConnectionFactory`?  Check how it is done with `SqlServerConnection`. Maybe it might make sense to simplify the connection creation in this PR to make it more understandable.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectionFactory.java
Naros: Great catch; I'm surprised that didn't somehow cause the connector to fail.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectionFactory.java
Naros: Should `USE_FETCH_SIZE_WITH_LONG_COLUMN` be defined in `JdbcConfiguration`?  It's not a known field and so what I saw is that its not being added as part of the default `JdbcConfiguration` object so I had to force add it if the configuration option wasn't being specified by the user's connector setup.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectionFactory.java
Naros: We can probably do that if I can understand the issue with the `USE_FETCH_SIZE_WITH_LONG_COLUMN` issue where it isn't part of the JdbcConfiguration and I have to force apply it when its not provided.

Naros:
@jpechane I'm actually doing a small rework to this as a little more testing tonight I stumbled across a corner case.  I'm going to do my best to get the rework in so you can review it tomorrow morning before the release if it all works out.  I'd definitely like to get this into Alpha1 if possible.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectionFactory.java
Naros: I removed this class entirely, I was able to leverage creating the factory similar to how SQL Server did.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectionFactory.java
Naros: You can ignore my comment above, I've removed this configuration option entirely so its a moot point.

jpechane:
@Naros Thanks a lot!

PR #205, opened 2020-10-21
DBZ-2683 Support partition tables

pikaaQ:
https://issues.redhat.com/browse/DBZ-2683

Partiioin tables' SEG_NAME is not 'TABLE_NAME', the 'SEG_NAME' condition rejected all partition table.

PR #206, opened 2020-11-02
DBZ-2683 Add support for partitioned tables

Naros:
https://issues.redhat.com/browse/DBZ-2683

I've taken PR #205 and pulled it into a separate PR with the unit test.  I did this because the original PR was opened against the contributor's master branch rather than a topic branch and didn't feel comfortable pushing the unit test to their fork that way.


jpechane:
Applied, thanks

PR #207, opened 2020-11-02
DBZ-2679 Handle LCR events with database names with domains

Naros:
https://issues.redhat.com/browse/DBZ-2679


jpechane:
@Naros LGTM, applied

PR #209, opened 2020-11-05
DBZ-2729 Implement Oracle LogMiner features/fixes

Naros:
https://issues.redhat.com/browse/DBZ-2729

This PR implements the following Oracle LogMiner changes:

1. Oracle RAC support
2. Historical Tracking
3. Better JMX metrics
4. Mining archive logs
5. Bugfixes & minor enhancements

@AndreyIg Can you please take a moment and review these changes that they look good to you?  I'll call-out a few specifics below that I'd like to further discuss explicitly.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
@@ -5,6 +5,7 @@
  */
 package io.debezium.connector.oracle;
 
+import java.util.Set;
 import java.util.function.Predicate;
 
 import org.apache.kafka.common.config.ConfigDef;
@@ -32,6 +33,7 @@ import io.debezium.relational.Tables;
 import io.debezium.relational.Tables.TableFilter;
 import io.debezium.relational.history.HistoryRecordComparator;
 import io.debezium.relational.history.KafkaDatabaseHistory;
+import io.debezium.util.Strings;
 
 import oracle.streams.XStreamUtility;
 
@@ -170,6 +172,37 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
             .withImportance(Importance.HIGH)
             .withDescription("A token to replace on snapshot predicate template");
 
+    public static final Field RECORD_LOG_MINING_HISTORY = Field.create("log.mining.record.history")
+            .withDisplayName("Record Log Mining history")
+            .withType(Type.BOOLEAN)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.HIGH)
+            .withDefault("false")
+            .withDescription("Flag to record Log Mining history");
+
+    public static final Field LOG_MINING_HISTORY_RETENTION = Field.create("database.history.retention.hours")
+            .withDisplayName("Log Mining history retention")
+            .withType(Type.STRING)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.MEDIUM)
+            .withDefault("4")
+            .withDescription("Hours to keep Log Mining history");
+
+    public static final Field RAC_SYSTEM = Field.create("database.rac")
+            .withDisplayName("Oracle RAC")
+            .withType(Type.BOOLEAN)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.HIGH)
+            .withDefault("false")
+            .withDescription("Flag to if it is RAC system");
+
+    public static final Field RAC_NODE_NAMES = Field.create("database.rac.nodes.ip")
+            .withDisplayName("Oracle RAC nodes ip")
+            .withType(Type.STRING)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.HIGH)
+            .withDescription("RAC nodes which are balanced to connect the application");
+
     /**
      * The set of {@link Field}s defined as part of this configuration.
      */
@@ -204,6 +237,10 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
             CONNECTOR_ADAPTER,
             LOG_MINING_STRATEGY,
             SNAPSHOT_ENHANCEMENT_TOKEN,
+            RECORD_LOG_MINING_HISTORY,
+            LOG_MINING_HISTORY_RETENTION,
+            RAC_SYSTEM,
+            RAC_NODE_NAMES,
             CommonConnectorConfig.EVENT_PROCESSING_FAILURE_HANDLING_MODE);
 
     private final String databaseName;
@@ -267,7 +304,7 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
                 CommonConnectorConfig.EVENT_PROCESSING_FAILURE_HANDLING_MODE);
         Field.group(config, "Connector", CommonConnectorConfig.POLL_INTERVAL_MS, CommonConnectorConfig.MAX_BATCH_SIZE,
                 CommonConnectorConfig.MAX_QUEUE_SIZE, CommonConnectorConfig.SNAPSHOT_DELAY_MS, CommonConnectorConfig.SNAPSHOT_FETCH_SIZE,
-                SNAPSHOT_ENHANCEMENT_TOKEN);
+                SNAPSHOT_ENHANCEMENT_TOKEN, RECORD_LOG_MINING_HISTORY, LOG_MINING_HISTORY_RETENTION, RAC_SYSTEM, RAC_NODE_NAMES);
 
         return config;
     }
@@ -641,6 +678,26 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
      */
     public LogMiningStrategy getLogMiningStrategy() {
         return LogMiningStrategy.parse(getConfig().getString(LOG_MINING_STRATEGY));
+    }
+
+    /**
+     * @return flag if mining history will be recorded
+     */
+    public Boolean recordLogMiningHistory() {
+        return getConfig().getBoolean(RECORD_LOG_MINING_HISTORY);
+    }
+
+    public long logMinerHistoryRetention() {
+        return Long.parseLong(getConfig().getString(LOG_MINING_HISTORY_RETENTION));
+    }
+
+    public Boolean isRacSystem() {
+        return getConfig().getBoolean(RAC_SYSTEM);
+    }
+
+    public Set<String> racNodeNames() {
+        String nodes = getConfig().getString(RAC_NODE_NAMES);
+        return Strings.setOf(nodes, String::new);
     }
 
     /**
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHelper.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHelper.java
@@ -5,9 +5,11 @@
  */
 package io.debezium.connector.oracle.logminer;
 
+import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.sql.CallableStatement;
 import java.sql.Connection;
+import java.sql.DriverManager;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
@@ -15,6 +17,7 @@ import java.sql.Statement;
 import java.sql.Timestamp;
 import java.time.Duration;
 import java.time.Instant;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -29,7 +32,10 @@ import org.slf4j.LoggerFactory;
 
 import io.debezium.connector.oracle.OracleConnection;
 import io.debezium.connector.oracle.OracleConnectorConfig;
+import io.debezium.jdbc.JdbcConfiguration;
 import io.debezium.jdbc.JdbcConnection;
+import io.debezium.util.Clock;
+import io.debezium.util.Metronome;
 
 /**
  * This class contains methods to configure and manage Log Miner utility
@@ -39,10 +45,35 @@ public class LogMinerHelper {
     private final static String UNKNOWN = "unknown";
     private final static Logger LOGGER = LoggerFactory.getLogger(LogMinerHelper.class);
 
-    private enum DATATYPE {
+    public enum DATATYPE {
         LONG,
         TIMESTAMP,
-        STRING
+        STRING,
+        FLOAT
+    }
+
+    private static Map<String, Connection> racFlushConnections = new HashMap<>();
+
+    static void instantiateFlushConnections(JdbcConfiguration config, Set<String> hosts) {
+        for (Connection conn : racFlushConnections.values()) {
+            if (conn != null) {
+                try {
+                    conn.close();
+                }
+                catch (SQLException e) {
+                    LOGGER.error("cannot close connection");
+                }
+            }
+        }
+        racFlushConnections = new HashMap<>();
+        for (String host : hosts) {
+            try {
+                racFlushConnections.put(host, createFlushConnection(config, host));
+            }
+            catch (SQLException e) {
+                LOGGER.error("cannot connect node {}", host);
+            }
+        }
     }
 
     /**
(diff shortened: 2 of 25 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHistoryRecorder.java
```diff
@@ -0,0 +1,296 @@
...
+
+    private final static Logger LOGGER = LoggerFactory.getLogger(LogMinerHistoryRecorder.class);
+
+    public LogMinerHistoryRecorder(LogMinerMetrics metrics, JdbcConfiguration configuration, long historyRetention) {
+        this.metrics = metrics;
+        this.configuration = configuration;
+        this.historyRetention = historyRetention;
+
+        try {
+            createExecutor();
```
Naros, line 73: @AndreyIg Is there a reason why an executor, separate thread & connection necessary for this recorder logic?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
```diff
@@ -237,4 +273,21 @@ private void updateStartScn() {
     public void commitOffset(Map<String, ?> offset) {
         // nothing to do
     }
+
+    // todo: brough tin for review, but has already been ported to error-handler retriable stuff
```
Naros, line 277: /all, ignore this block, I'm going to be removing this section, I added this for my tracking some specific connector retry handling logic that was done one way on the older contributor's fork to how we do this upstream.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
```diff
@@ -129,6 +328,7 @@ private static String buildTableInPredicate(List<String> tables) {
 
         StringJoiner tableNames = new StringJoiner(",");
         tables.forEach(table -> tableNames.add("'" + table + "'"));
+        // todo: discuss with Andrey about partition tables here
```
Naros, line 331: @AndreyIg This was specifically shown by an upstream contributor to be problematic for partition tables, which we've since changed.  I wanted to highlight this in case you wanted to pull this in in your fork for now.

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/LogMinerHelperIT.java
```diff
@@ -0,0 +1,229 @@
...
+import io.debezium.connector.oracle.util.TestHelper;
+import io.debezium.embedded.AbstractConnectorTest;
+import io.debezium.util.Testing;
+
+/**
+ * This subclasses common OracleConnectorIT for LogMiner adaptor
+ */
+@SkipWhenAdapterNameIsNot(value = SkipWhenAdapterNameIsNot.AdapterName.LOGMINER, reason = "LogMiner specific tests")
+@Ignore(value = "The super class before method fails, needs investigation - alter session insufficient privileges")
+public class LogMinerHelperIT extends AbstractConnectorTest {
```
Naros, line 43: @AndreyIg As shown here, I've marked this test as `@Ignore`.  The problem with this test is I'm getting some insufficient privileges when the `@BeforeClass` handler is called and executes the `#resetSessionToCdb`.  I need to investigate why this is problematic here.

jpechane:
@Naros What is your plan in this regard? If approved by @AndreyIg are we goint to merge this one and then work on follow-up change/improvements?

Naros:
@jpechane Personally, I'd like to get this into the next 1.4 alpha release.  The archive log support I believe is critical for users who may need to restart the connector and potentially have a small down window, which the current implementation doesn't work well within that scope.  Getting this into the hands of users gives us a chance to get feedback on these sooner rather than later, which I'd rather do than repeat what happened in 1.3 Final.  @AndreyIg do you have some cycles to verify what is here looks good as well as address my open questions?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerMetricsMXBean.java
```diff
@@ -104,4 +104,55 @@
...
+    /**
+     * this flag turns on recording of captured incremental changes. It creates an overhead on CPU and takes disk space
+     * @param doRecording true - record
+     */
+    void setRecordMiningHistory(boolean doRecording);
+
+    // record log mining history flag
+    boolean getRecordMiningHistory();
+
+    // *** following metrics work if RecordMiningHistory is true.
```
Naros, line 119: Since the goal is to provide the `HistoryRecorder` as a hook rather than an implementation upstream, I think these metrics might make sense that they be exposed elsewhere as part of another JMX bean.  @jpechane would you agree?

Naros (commented):
See more comments inline.  I'm going to open a draft supplemental PR that is based on this with some upcoming changes as I don't want to taint these for review purposes by @AndreyIg.  I can then overlay my changes after his review.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerMetricsMXBean.java
Naros: After reviewing these metrics, I'm curious why these are never actually used in the runtime code, @AndreyIg?  I see we have a test case that checks that the method calls for these manipulate the metrics state accordingly; however, we actually never track these at runtime.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHistoryRecorder.java
AndreyIg: The main idea to keep this logic separate. If something happens, like tablespace cannot be extended, we stop recording with no affect on the main logic.
It also has separate queue.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
AndreyIg: Noted, good point. our customers have no partitioned tables yet

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/LogMinerHelperIT.java
AndreyIg: Please share your findings.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -643,6 +680,26 @@ public LogMiningStrategy getLogMiningStrategy() {
...
 
+    /**
+     * @return flag if mining history will be recorded
+     */
+    public Boolean recordLogMiningHistory() {
+        return getConfig().getBoolean(RECORD_LOG_MINING_HISTORY);
+    }
+
+    public long logMinerHistoryRetention() {
+        return Long.parseLong(getConfig().getString(LOG_MINING_HISTORY_RETENTION));
```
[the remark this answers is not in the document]
AndreyIg: please do. Thank you for the finding.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -170,6 +172,37 @@
...
+
+    public static final Field RAC_SYSTEM = Field.create("database.rac")
+            .withDisplayName("Oracle RAC")
+            .withType(Type.BOOLEAN)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.HIGH)
+            .withDefault("false")
+            .withDescription("Flag to if it is RAC system");
+
+    public static final Field RAC_NODE_NAMES = Field.create("database.rac.nodes.ip")
```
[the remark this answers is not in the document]
AndreyIg: host names did not work for me, IP worked. It could be a consequence or not.
We had no time to re-check. Please consider it as IPs

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHelper.java
```diff
@@ -272,10 +420,8 @@ static void checkSupplementalLogging(OracleConnection connection, String pdbName
                 connection.setSessionToPdb(pdbName);
             }
 
-            final String key = "KEY";
-            String validateGlobalLogging = "SELECT '" + key + "', " + " SUPPLEMENTAL_LOG_DATA_ALL from V$DATABASE";
-            Map<String, String> globalLogging = getMap(connection.connection(false), validateGlobalLogging, UNKNOWN);
-            if ("no".equalsIgnoreCase(globalLogging.get(key))) {
+            Map<String, String> globalLogging = getMap(connection.connection(false), SqlUtils.supplementalLoggingCheckQuery(), UNKNOWN);
```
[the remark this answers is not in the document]
AndreyIg: You are absolutely correct. Ideally we have to configure table-level groups for whitelisted columns.
We did it on database level , because of two reasons:
- we wanted to automate it.  Initially I run table level DDLs from the connector code,:
        if (pdbName != null) {
            jdbcConnection.setSessionToPdb(pdbName);
        }

        final String key = "KEY";
        String validateGlobalLogging = "SELECT '" + key + "', " + " SUPPLEMENTAL_LOG_DATA_ALL from V$DATABASE";
        Map<String, String> globalLogging = getMap(connection, validateGlobalLogging, "unknown");
        if ("no".equalsIgnoreCase(globalLogging.get(key))) {
            tableIds.forEach(table -> {
                String tableName = table.schema() + "." + table.table();
                try {
                    String validateTableLevelLogging = String.format("SELECT '%s', LOG_GROUP_TYPE FROM DBA_LOG_GROUPS WHERE LOG_GROUP_TYPE='ALL COLUMN LOGGING' AND OWNER ='%s' AND TABLE_NAME = '%s'", key,
                            table.schema().toUpperCase(), table.table().toUpperCase());
                    Map<String, String> tableLogging = getMap(connection, validateTableLevelLogging, "unknown");
                    if (tableLogging.get(key) == null) {
                        String alterTableStatement = "ALTER TABLE " + tableName + " ADD SUPPLEMENTAL LOG DATA (ALL) COLUMNS";
                        LOGGER.info("altering table {} for supplemental logging", table.table());
                        executeCallableStatement(connection, alterTableStatement);
                    }
                } catch (SQLException e) {
                    throw new RuntimeException("Cannot set supplemental logging for table " + tableName, e);
                }
            });
        } else {
            LOGGER.warn("Supplemental logging is set on global level, setting individual table supplemental logging was skipped");
        }

        if (pdbName != null) {
            jdbcConnection.resetSessionToCdb();
        }

 but logminer did not see changes from those tables. I cannot recall why.
- we had no time to experiment with column groups and I set it on the global level for the deadline sake.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHistoryRecorder.java
```diff
@@ -0,0 +1,296 @@
...
+                            }
+                            catch (SQLException e1) {
+                                LOGGER.error("Retry failed: {}", e1);
+                            }
+                        }
+                    }
+                    LOGGER.error("Cannot insert data due to {}. " +
+                            "SCN: {}, Table name: {}, Segment Owner: {}, operation code: {}, ChangeTime: {}, transaction id: {}, csf: {}, sql: {}",
+                            e, scn.longValue(), tableName, segOwner, operationCode, changeTime, transactionId, csf, redoSql);
+                    // todo remove after figuring out what the problem is
```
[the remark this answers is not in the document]
AndreyIg: there is an Oracle bug: Doc ID 2620584.1 which recommends a workaround. Instead re just retry.
This is happening once in a few tens of millions of records. I cannot come up with a better solution.
This is not critical, because we would have REDO_SQL in the log and will continue.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHistoryRecorder.java
```diff
@@ -0,0 +1,296 @@
...
+ * 2. Replaying recorded history
+ *    - for regression testing. This will not cover all application part, like parsing, because we record already parsed redoSql.
+ *      it could be beneficial for consumer testing for instance
+ *    - if Kafka is down (further code modification needed)
+ *    - for database migration. Do consistent dump using DPDUMP utility and replay the tail from this history table.
+ * todo ensure thread safeness for multi-threaded parser branch
+ * todo allow dynamic initialization after resolving DB problems
+ */
+@NotThreadSafe
+public class LogMinerHistoryRecorder implements HistoryRecorder, AutoCloseable {
```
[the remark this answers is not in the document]
AndreyIg: you are right and we discussed it already. History recording could be implemented on the local logging level in the future

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerQueryResultProcessor.java
```diff
@@ -196,44 +201,50 @@ int processResult(ResultSet resultSet) {
...
-     * The criteria is the offset SCN remains the same in five mining cycles
+     * The criteria is the offset SCN remains the same in 25 mining cycles
      */
     private void warnStuckScn() {
         if (offsetContext != null && offsetContext.getCommitScn() != null) {
             if (currentOffsetScn == offsetContext.getScn() && currentOffsetCommitScn != offsetContext.getCommitScn()) {
                 stuckScnCounter++;
                 // logWarn only once
-                if (stuckScnCounter == 5) {
+                if (stuckScnCounter == 25) {
```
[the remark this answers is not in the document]
AndreyIg: I believe it could be totally eliminated. We have added this just to simplify troubleshooting. With history recording this is irrelevant.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerMetricsMXBean.java
AndreyIg: agree, we should eliminate those

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHistoryRecorder.java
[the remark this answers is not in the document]
jpechane: @Naros +1 as you did in the other draft PR

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerMetricsMXBean.java
jpechane: @Naros +1

Naros:
Thanks for the feedback @AndreyIg, I'll adjust the follow-up PR #213 based on your input here.  If there are any outstanding questions, I'll log a follow-up jira related to them and we'll address those in the next alpha/beta release in a few weeks.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHistoryRecorder.java
Naros: Sounds good.  Since I've removed `LogMinerHistoryRecorder` implementation from upstream Debezium, how you intend to manage the process is entirely now user-specific.  As we move forward, if there are additional hooks needed on the `HistoryRecorder` interface, let us know.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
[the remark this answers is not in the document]
Naros: I've indicated the field to be either host names or IP addresses.  Both should work in theory since the field is meant to be a comma-separated list of values.  One thing to note for you downstream is that we've done some consolidation of configuration options introduced by your code. 

For Oracle RAC, we've removed the boolean toggle in favor of just determining whether Oracle RAC is enabled based on if the comma-separated value-set for `rac.nodes` is empty or not.  Additionally, we did something similar with history recording where we determine whether its enabled based on if the retention for history recording is greater than `0`.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHelper.java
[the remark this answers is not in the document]
Naros: I think its reasonable to expect that a user has configured the table-level supplemental logging prior to connector start.  I have an open JIRA to look at this, which I plan to investigate this coming week before holiday.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHistoryRecorder.java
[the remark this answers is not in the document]
Naros: No problem.  This is no longer a concern for us with the `HistoryRecorder` API refactor.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerQueryResultProcessor.java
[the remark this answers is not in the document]
Naros: I'm going to leave this as-is and take a second pass at this after 1.4.0.Alpha2.  I need to understand for the use case where history recording isn't enabled (the default), would this be necessary or provide any benefit.

Naros:
Closing in favor of PR #213.

PR #210, opened 2020-11-10
DBZ-1720 Move CI to GH Actions

ani-sha:


.github/workflows/build.yml
```diff
@@ -0,0 +1,40 @@
...
+        uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+      - name: Setup Debezium Parent
+        run: git clone https://github.com/debezium/debezium.git $PWD/core-repo && cd $PWD/core-repo && mvn -B clean install -DskipTests -DskipITs -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn && cd $OLDPWD
+      - name: Setup Debezium Incubator
+        run: mvn -B clean install -Passembly -Dformat.formatter.goal=validate -Dformat.imports.goal=check -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
```
Naros, line 31: I'm curious for consistency if we shouldn't look at creating multiple workflows much like we did in the core repository where we can control which connectors get built depending on the changed files of a push or pull request?  We don't build Oracle on CI presently, so we'd really only need to focus on a cassandra and db2 workflow presently.  wdyt?

ani-sha: That's a better one I guess, will be adding separate workflows for `cassandra` and `db2`.

jpechane:
@ani-sha Nice work! Could you please add also formatting and commit message workflows from core repo?

ani-sha:
@jpechane Thanks! Should I add a separate PR for that or add it here itself?

jpechane:
@ani-sha The followup as separate PR. The rest in this PR.

ani-sha:
> @ani-sha The followup as separate PR. The rest in this PR.

Sure! :100:

.github/workflows/build.yml
```diff
@@ -0,0 +1,36 @@
+name: Build Debezium Incubator
```
jpechane, line 1: @ani-sha I belive this is completely unnecessary as these are steps specific for Travis only. Maven is already present in the ubuntu-latest env.

Also I belive that the `cache` shoudl be present on connector workflows as this is the place where the cache is needed.

ani-sha: @jpechane Best to remove this workflow as we don't need the `webhooks` too. And adding the `cache` to connector workflows.

jpechane:
@ani-sha Applied, thanks!

PR #211, opened 2020-11-10
DBZ-1446 Avoid Oracle "no snapshot found based on specified time" during snapshot

Naros:
https://issues.redhat.com/browse/DBZ-1446

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
@@ -187,10 +187,10 @@ public class OracleSnapshotChangeEventSource extends RelationalSnapshotChangeEve
             return Optional.of(rs.getLong(1));
         }
         catch (SQLException e) {
-            if (e.getMessage().startsWith("ORA-08180")) {
+            if (e.getErrorCode() == 8180) {
                 // DBZ-1446 In this use case we actually do not want to propagate the exception but
                 // rather return an empty optional value allowing the current SCN to take prior.
-                LOGGER.debug("No latest table SCN could be resolved, defaulting to current SCN");
+                LOGGER.info("No latest table SCN could be resolved, defaulting to current SCN");
                 return Optional.empty();
             }
             throw e;
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
@@ -186,6 +186,15 @@ public class OracleSnapshotChangeEventSource extends RelationalSnapshotChangeEve
 
             return Optional.of(rs.getLong(1));
         }
+        catch (SQLException e) {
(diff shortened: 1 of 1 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
```diff
@@ -186,6 +186,15 @@ private boolean areSameTimestamp(Long scn1, long scn2) throws SQLException {
 
             return Optional.of(rs.getLong(1));
         }
+        catch (SQLException e) {
+            if (e.getMessage().contains("no snapshot found based on specified time")) {
```
Perdjesk, line 190: Wouldn't it better to match against the ORA error code (i.e ORA-08180)? 
Asking as well whether there is no a better/programmatically manner to check for this cases instead of relying on error string matching.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
```diff
@@ -186,6 +186,15 @@ private boolean areSameTimestamp(Long scn1, long scn2) throws SQLException {
...
+                // DBZ-1446 In this use case we actually do not want to propagate the exception but
+                // rather return an empty optional value allowing the current SCN to take prior.
+                LOGGER.warn("No latest table SCN could be resolved, defaulting to current SCN");
```
Perdjesk, line 193: Not sure it is enough important to be a warning, given that can be a common and expected case and that in such case it is safe to trust the current SCN.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
Naros: We can definitely match it against `startsWith("ORA-08180")`.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
Naros: That's a good point, I'll change it to `debug` instead since it really doesn't have any major impact in normal operation.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
jpechane: @Naros As we match against the error code now I updated to code to use error code from the exception isnated of message

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
jpechane: @Naros I've elevated the level to info. The message happens only once and it can be useful information for issue diagnostics in case there are some unlanned sideeffects.

jpechane:
@Naros Apllied, thanks

Naros:
@jpechane Thanks for the self corrections :)

PR #214, opened 2020-11-12
DB2 Connector - Table and field name using lowercase

muswanto:
I've a question, how to setting db2 connector when my tabel name and fields using lower case?
I can capture it when table name is uppercase. Please advice, thank you


jpechane:
@muswanto Hi please use our chat or mailing list for questions. When you do it please share your connector configuration and table/dba name you fail to capture.

muswanto:
Hi @jpechane, thank you for reply, but where're chat or mailing list? I'll post this on there.

PR #215, opened 2020-11-13
DBZ-2755

zrlurb:
Using CURRENT SERVER special register instead of CURRENT_SERVER?


jpechane:
@zrlurb Could you please update the commit message to be like `DBZ-2755 Use CURRENT SERVER query` or similar?

jpechane:
@zrlurb Applied thanks! I've just updated the log message

PR #232, opened 2021-01-06
DBZ-2890 Remove SLF4J from Oracle distro

jpechane:
https://issues.redhat.com/browse/DBZ-2890


gunnarmorling:
Good spot; do we need  this for 1.4, too?

PR #235, opened 2021-01-18
DBZ-2920 Fix Logminer DML operations using continuation

Naros:
https://issues.redhat.com/browse/DBZ-2920


gunnarmorling:
Applied; thx, @Naros!

PR #230, opened 2020-12-25
DBZ-2711 Supplemental logging is required for entire database rather than per monitored table

ant0nk:
I've tested snapshotting and replication with only SUPPLEMENTAL_LOG_DATA_MIN enabled - all works same as with SUPPLEMENTAL_LOG_DATA_ALL.

PR #238, opened 2021-01-22
DBZ-2711 Implement relaxed supplemental logging requirements for Logminer

Naros:
https://issues.redhat.com/browse/DBZ-2711

This PR builds atop of PR #230 by not only expecting minimal database level supplemental logging but also adding a check to verify that tables which are monitored by Oracle specify all columns being logged to maintain consistency with old behavior.


gunnarmorling:
Thx, @ant0nk and @Naros!

PR #239, opened 2021-01-26
DBZ-2855: Missing log file error when current SCN is very large

mpermar:
This PR fixes a problem that happens when the NEXT_CHANGE# column in the [V$LOG](https://docs.oracle.com/en/database/oracle/oracle-database/19/refrn/V-LOG.html#GUID-FCD3B70B-7B98-40D8-98AB-9F6A85E69F57) view is longer than what the Long data type supports in Java which is a signed long whereas the Oracle value is an unsigned long. 

This PR uses BigInteger as an alternative approach to hold such large numbers. Alternatively, we could use the unsigned long support in Java 8 ( Long.parseUnsignedLong). I think that would be much better but I'm not aware of the backwards compatibility policy of this project so leaving that choice up to the team. 

Nevertheless, I believe some additional JIRA and PR work is needed to guarantee there are no other similar issues within the rest of the source code and migrate it to use BigInteger or unsigned longs where necessary.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHelper.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHelper.java
@@ -44,7 +44,7 @@ public class LogMinerHelper {
     private static final String TOTAL = "TOTAL";
     private final static Logger LOGGER = LoggerFactory.getLogger(LogMinerHelper.class);
 
-    public final static String MAX_SCN_S = "1844674407370955161";
+    public final static String MAX_SCN_S = "18446744073709551615";
     public final static BigInteger MAX_SCN_BI = new BigInteger(MAX_SCN_S);
 
     public enum DATATYPE {
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHelper.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHelper.java
@@ -5,7 +5,6 @@
  */
 package io.debezium.connector.oracle.logminer;
 
-import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.sql.CallableStatement;
 import java.sql.Connection;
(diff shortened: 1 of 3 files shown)


Naros:
> Nevertheless, I believe some additional JIRA and PR work is needed to guarantee there are no other similar issues within the rest of the source code and migrate it to use BigInteger or unsigned longs where necessary.

@mpermar Can you confirm that you're operating on Oracle 19c?  I did some documentation digging on `V$LOG` specifically and it seems like for both Oracle 11 and 12, the max value one would expect is `281474976710655` while the documentation for 19c shows to expect a maximum value of `18446744073709551615` if the status of the entry is set to `CURRENT`.   

With that said, under versions prior to Oracle 19, that would mean we would never hit this scenario as the long data type would be large enough to support these values; however with Oracle 19 that's no longer the case.  But I also don't think any system's SCN with Oracle 19 has reached a point to mandate full `BigInteger` compliance yet most likely and we're just hitting this corner case when the status is `CURRENT` and the extraordinarily large value is being seeded by Oracle.

If we agree that's likely the corner case here, I'm fine with applying these changes directly so we can get Logminer working with Oracle 19c to avoid this corner case and I'll add a follow-up PR to where we look at Oracle 19c from a larger compliance & support perspective.  

@gunnarmorling @jpechane any objections?

mpermar:
@Naros , yes, I'm running on 19c.

Naros:
> yes, I'm running on 19c.

Thanks for confirming @mpermar and is it safe to assume in your environment you've deployed this hotfix locally and you haven't seen anymore `missing log file` errors to your knowledge?

mpermar:
I haven't observed any issues when restarting the connector.

Naros:
Awesome news, I'll take these changes and run them against Oracle 12 for compatibility and assuming tests pass I'll proceed with getting this merged so we can include it in 1.4.1.Final tomorrow.

Naros:
I've logged [DBZ-2994](https://issues.redhat.com/browse/DBZ-2994) as a follow-up, all tests pass with Oracle 12c with these changes so I'm going to proceed with merging this as-is for now.

PR #241, opened 2021-01-26
DBZ-2981 Debezium swallows DML exception in certain cases

mpermar:
Very simple fix. SLF4J handles the last argument specially if it is an exception so that will get the stacktrace printed. This helps with exceptions like a NullPointerException as otherwise all we would get is a simple "null".


Naros (approved):
LGTM

PR #242, opened 2021-01-28
DBZ-2891 Oracle Xstream deadlock work

Naros:
https://issues.redhat.com/browse/DBZ-2891

@jpechane This PR isn't yet ready, this contains just the integration test that I'm using to try and find a solution.  So far I haven't been able to devise a solution using a type of lock/flag in order to avoid the deadlock reliably.  I'll try to circle back with you when you arrive in the morning to discuss next steps.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/XstreamStreamingChangeEventSource.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/XstreamStreamingChangeEventSource.java
@@ -5,6 +5,7 @@
  */
 package io.debezium.connector.oracle.xstream;
 
+import java.time.Duration;
 import java.util.Map;
 
 import org.slf4j.Logger;
@@ -20,9 +21,13 @@ import io.debezium.pipeline.EventDispatcher;
 import io.debezium.pipeline.source.spi.StreamingChangeEventSource;
 import io.debezium.relational.TableId;
 import io.debezium.util.Clock;
+import io.debezium.util.Metronome;
 
 import oracle.jdbc.OracleConnection;
 import oracle.sql.NUMBER;
+import oracle.streams.ChunkColumnValue;
+import oracle.streams.LCR;
+import oracle.streams.RowLCR;
 import oracle.streams.StreamsException;
 import oracle.streams.XStreamOut;
 import oracle.streams.XStreamUtility;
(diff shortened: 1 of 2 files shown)


Naros:
@jpechane This is ready for review.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/XstreamStreamingChangeEventSource.java
```diff
@@ -139,4 +159,23 @@ else if (scn != null) {
...
+        // This is to mimic the behavior of the Xstream callback API
+        if (lcr instanceof RowLCR) {
+            if (((RowLCR) lcr).hasChunkData()) {
+                ChunkColumnValue chunk;
+                do {
+                    chunk = xsOut.receiveChunk(XStreamOut.DEFAULT_MODE);
+                    handler.processChunk(chunk);
+                } while (!chunk.isEndOfRow());
+            }
+        }
```
Naros, line 179: We could debate whether this is even necessary given the fact that the callback handler isn't written to support chunks but I opted to include this mainly to remain consistent with what would have be observed by using the callback handler previously if we had encountered a row that had chunk data.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/XstreamStreamingChangeEventSource.java
```diff
@@ -71,11 +76,26 @@ public void execute(ChangeEventSourceContext context) throws InterruptedExceptio
...
+                    // When a batch begins, Xstream expects that to end before a new operation can
+                    // be performed, which includes detachment. This is why we only test for context
+                    // shutdown once we've left the LCR batch execution status.
+                    if (!context.isRunning()) {
+                        break;
+                    }
+
+                    // Xstream callback API implies a 1 second delay between batches
+                    // To mimic that behavior with the non-callback API, we apply a pause of 1 second
+                    metronome.pause();
```
gunnarmorling, line 97: So to be clear, is this something that happens "often"? Or only rarely? I.e. is there any impact on latency in the common case? Or would it have been the same with the previous approach, too?

Naros: So the callback API, e.g. `receiveLCRCallback` internally applies a 1 second delay between batches, and goal was to keep that the same.  This meant that if an LCR was returned, we wouldn't apply the pause until we had drained all the LCR records from Xstreams and we would then apply a 1 second poll until a new record was detected.

Naros:
Closing this PR in favor of #243.  I'll leave this branch around in case we need to resurrect the changes.

PR #243, opened 2021-01-28
DBZ-2891 Set watermark in the XStream thread

jpechane:
https://issues.redhat.com/browse/DBZ-2891

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
@@ -5,8 +5,6 @@
  */
 package io.debezium.connector.oracle.xstream;
 
-import java.util.concurrent.atomic.AtomicReference;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -18,6 +16,7 @@ import io.debezium.pipeline.EventDispatcher;
 import io.debezium.relational.RelationalDatabaseSchema;
 import io.debezium.relational.TableId;
 import io.debezium.util.Clock;
+
 import oracle.streams.ChunkColumnValue;
 import oracle.streams.DDLLCR;
 import oracle.streams.LCR;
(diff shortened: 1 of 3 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/XstreamStreamingChangeEventSource.java
```diff
@@ -99,35 +101,18 @@ public void execute(ChangeEventSourceContext context) throws InterruptedExceptio
...
                 LOGGER.debug("Recording offsets to Oracle");
                 final LcrPosition lcrPosition = LcrPosition.valueOf((String) offset.get(SourceInfo.LCR_POSITION_KEY));
                 final Long scn = (Long) offset.get(SourceInfo.SCN_KEY);
-                if (lcrPosition != null) {
-                    if (LOGGER.isDebugEnabled()) {
-                        LOGGER.debug("Recording position {}", lcrPosition);
+                final PositionAndScn position = new PositionAndScn(lcrPosition, scn);
+                for (int i = 0; (i < 10) && !lcrMessage.compareAndSet(null, position); i++) {
+                    try {
+                        Thread.sleep(100);
```
gunnarmorling, line 110: This is pretty concerning.

jpechane: That's intended - of course would be written with our utils. The key idea is that we can affored to lose few flushes. As soon as the message queue is unblocked a flush will be executed. In this case it might be even unnecessary to do that. As I think about that we can overite it always. As the LCR handler is ublkock it will flush the highest position.

jpechane:
@gunnarmorling This is ready for final review

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
```diff
@@ -37,20 +41,28 @@
     private final RelationalDatabaseSchema schema;
     private final OracleOffsetContext offsetContext;
     private final boolean tablenameCaseInsensitive;
+    private final AtomicReference<PositionAndScn> lcrMessage;
```
gunnarmorling, line 44: I'd suggest to not have the atomic reference here directly, but rather have a method `PositionAndScn getAndResetPublishedPosition()` on `XstreamStreamingChangeEventSource`.

jpechane: @gunnarmorling Done, I also added send message for symmetry

gunnarmorling (approved):

gunnarmorling:
Applied; thx!

PR #246, opened 2021-01-29
DBZ-2982 Allow to customize log mining properties

mpermar:
This change makes some of the log mining constant properties available to be tuned through the connector configuration. I've added support for the SCN batch window properties as well as for sleep time and fetch size.

This change alleviates slightly the issue reported on DBZ-2982 in the sense that it provides some capability to increase the batch window and sleep time so in case of encountering a SCN jump it will be possible to soften the effects by using larger batch sizes and smaller sleep intervals. Of course this might not target some of the use cases and ideally this PR should be complemented by additional improvements like either a better algorithm or a different approach to handle those SCN jumps better.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
@@ -50,6 +50,17 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
     public static final String DATABASE_CONFIG_PREFIX = "database.";
 
     protected static final int DEFAULT_PORT = 1528;
+    
+    protected static final int DEFAULT_VIEW_FETCH_SIZE = 10_000;
+
+    protected final static int DEFAULT_BATCH_SIZE = 20_000;
+    protected final static int MIN_BATCH_SIZE = 1_000;
+    protected final static int MAX_BATCH_SIZE = 100_000;
+
+    protected final static int MAX_SLEEP_TIME = 3_000;
+    protected final static int DEFAULT_SLEEP_TIME = 1_000;
+    protected final static int MIN_SLEEP_TIME = 0;
+    protected final static int SLEEP_TIME_INCREMENT = 200;
 
     public static final Field HOSTNAME = Field.create(DATABASE_CONFIG_PREFIX + JdbcConfiguration.HOSTNAME)
             .withDisplayName("Hostname")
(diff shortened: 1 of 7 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerMetrics.java
```diff
@@ -90,7 +103,7 @@ public void reset() {
         totalBatchProcessingDuration.set(Duration.ZERO);
         maxBatchProcessingThroughput.set(0);
         lastBatchProcessingDuration.set(Duration.ZERO);
-        networkConnectionProblemsCounter.set(0);
+        networkConnectionProblemsCounter.set(0);
```
Naros, line 106:
```suggestion
        networkConnectionProblemsCounter.set(0);
```
Lets avoid the trailing spaces.

Naros (commented):
Overall this LGTM.  

I did have one idea I'd like to consider including and that would be to log something to the actual connector log when the adaptive batch window shifts.  The idea is that if users mention that it looks like the connector isn't working but it is, logging something would make diagnosing this use case far easier from outside looking in.

The big question is what log level should this be written to.  My gut says we probably should do it as debug; however, there could be merit to doing this as info if and only if its infrequent enough that it wouldn't cause considerable log bloat.  Since you've had more experience with this situation in your environment; how frequent is this?  If its frequent, I'm fine with using debug level.

Additionally, this PR needs one additional PR opened against the main Debezium repository so that we document these new options in the `oracle.adoc` documentation.  Would you be willing to do this @mpermar ?  

Lastly, would you be willing to share your connector custom values for these in the associated Jira?  I'd like to add a section to the Oracle docs to discuss when things seem or do go wrong with the connector and I'd like to use your values as a basis to illustrate an example in such a section in the documentation.

mpermar:
@Naros yes, let me send a PR for the docs

On the logging. The way batching is implemented I believe it changes with every iteration. Hence +1 on debug. I have pushed a new commit that essentially will resort to info when we get to the max batch size and otherwise just do the debug.

mpermar:
@Naros here is the related documentation PR: https://github.com/debezium/debezium/pull/2085

gunnarmorling:
@mpermar, could you please rework the commits to use an email address associated to your GH account, and also rebase it (there shouldn't be any merge commits)? Thx!

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -222,6 +233,75 @@
...
+    public static final Field LOG_MINING_SLEEP_TIME_MIN = Field.create("log.mining.sleep.time.min")
+            .withDisplayName("Minimum sleep time in milliseconds when reading redo/archive logs.")
+            .withType(Type.LONG)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.LOW)
+            .withDefault(MIN_SLEEP_TIME)
+            .withDescription(
+                    "The minimum amount of time that the connector will sleep after reading data from redo/archive logs and before starting reading data again. Value is in milliseconds.");
+
+    public static final Field LOG_MINING_SLEEP_TIME_DEFAULT = Field.create("log.mining.sleep.time.default")
```
gunnarmorling, line 278: All temporal option names should be suffixed with ".ms".

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -715,6 +803,70 @@ public Duration getLogMiningArchiveLogRetention() {
...
+     */
+    public int getLogMiningBatchSizeDefault() {
+        return getConfig().getInteger(LOG_MINING_BATCH_SIZE_DEFAULT);
+    }
+
+    /**
+     * 
+     * @return int The minimum sleep time used when mining redo/archive logs
+     */
+    public int getLogMiningSleepTimeMin() {
```
gunnarmorling, line 842: Should return `java.time.Duration` (same below).

gunnarmorling (commented):
This looks like a good improvement, thanks a lot, @mpermar! A few small comments inline.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
mpermar: Sure

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
mpermar: Ah good point. I hadn't spotted the pattern.

mpermar:
@gunnarmorling , @Naros , updated the PR.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
Naros: I noticed a commit description that made reference to suffixes; however I don't see anywhere in the PR where this was addressed so that these configuration options match the property names in the supplemental docs.  Is this still in the works?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
Naros: Like with the suffix comment above, I see the latest commit suggests the usage of `Duration` but the commit doesn't appear to include such changes.

mpermar: 🤦‍♂️
Should have looked within the commit. Proper change was done on a different local fork yet of course the commit filenames matched. Anyways, pebkac, updated PR with the proper renaming and typing.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerMetrics.java
```diff
@@ -235,7 +235,7 @@ public int getBatchSize() {
 
     @Override
     public Integer getMillisecondToSleepBetweenMiningQuery() {
-        return millisecondToSleepBetweenMiningQuery.get();
+        return millisecondToSleepBetweenMiningQuery.intValue();
```
Naros, line 238: I would suggest we actually return the value using `get()` here and set the return type as `long` as there is no need to return the value as `Integer` or even `int` when all the usages expect `long` anyway.
```suggestion
        return millisecondToSleepBetweenMiningQuery.get();
```

Naros: I'll make the above change myself when I commit these changes; otherwise this LGTM.

mpermar: 👍 I decided to go with intValue() assuming the override was in the parent class from debezium core. I see now it is in the jmx interface.

Naros:
I've resolved the conflicts & merged into master.  Thanks @mpermar!

PR #248, opened 2021-02-03
DBZ-2420 Use database configs from RelationalDatabaseConnectorConfig

ani-sha:
https://issues.redhat.com/browse/DBZ-2420

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
@@ -54,6 +54,10 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
     public static final Field PORT = RelationalDatabaseConnectorConfig.PORT
             .withDefault(DEFAULT_PORT);
 
+    public static final Field HOSTNAME = RelationalDatabaseConnectorConfig.HOSTNAME
+            .withNoValidation()
+            .withValidation(OracleConnectorConfig::requiredWhenNoUrl);
+
     public static final Field PDB_NAME = Field.create(DATABASE_CONFIG_PREFIX + "pdb.name")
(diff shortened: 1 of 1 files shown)


Naros:
These changes presently cause `OracleConnectorConfigTest` to fail; investigating the failures.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -51,46 +51,8 @@
 
     protected static final int DEFAULT_PORT = 1528;
 
-    public static final Field HOSTNAME = Field.create(DATABASE_CONFIG_PREFIX + JdbcConfiguration.HOSTNAME)
-            .withDisplayName("Hostname")
-            .withType(Type.STRING)
-            .withWidth(Width.MEDIUM)
-            .withImportance(Importance.HIGH)
-            .withValidation(OracleConnectorConfig::requiredWhenNoUrl)
```
Naros, line 59: I don't believe we can use the `RelationalDatabaseConnectorConfig` definition as-is since the Oracle implementation has a special validation that allows the `database.hostname` to be omitted when a URL is provided.  

```suggestion
    public static final Field HOSTNAME = RelationalDatabaseConnectorConfig.HOSTNAME
            .withNoValidation()
            .withValidation(OracleConnectorConfig::requiredWhenNoUrl);          
```

The remaining references to `RelationalDatabaseConnectorConfig.HOSTNAME` should then be changed to use the connector specific HOSTNAME reference instead.

Naros: I added a commit that satisfies this requirement, tests now all pass.

Naros:
Thanks @ani-sha!

ani-sha:
Thanks a lot @Naros for the review and changes. :pray:

PR #253, opened 2021-02-11
DBZ-3078 Initial implementation of FastDmlParser

Naros:
https://issues.redhat.com/browse/DBZ-3078

This PR is still a WIP as we want to introduce a configurable option to toggle back to the legacy parser & the code itself needs to be adjusted to use this parser implementation.


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/FastDmlParser.java
```diff
@@ -0,0 +1,472 @@
...
+import io.debezium.connector.oracle.logminer.valueholder.LogMinerColumnValue;
+import io.debezium.connector.oracle.logminer.valueholder.LogMinerColumnValueImpl;
+import io.debezium.connector.oracle.logminer.valueholder.LogMinerDmlEntry;
+import io.debezium.connector.oracle.logminer.valueholder.LogMinerDmlEntryImpl;
+import io.debezium.data.Envelope;
+
+/**
+ * @author Chris Cranford
+ */
+public class FastDmlParser {
```
gunnarmorling, line 22: LOL, no pressure there ;)

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/FastDmlParser.java
```diff
@@ -0,0 +1,472 @@
...
+    private static final String VALUES = " values ";
+
+    /**
+     * Parse a DML SQL statement.
+     *
+     * @param sql the sql statement
+     * @return the parsed DML entry record or {@code null} if the SQL was not parsed
+     */
+    public LogMinerDmlEntry parse(String sql) {
+        if (sql.startsWith(INSERT_INTO)) {
```
gunnarmorling, line 42: Could be sql.charAt(0) == 'i' perhaps?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/FastDmlParser.java
Naros: Call it as I see it :)

Naros:
I applied your suggested change as well as renamed the parser to `LogMinerDmlParser`.  The converters were also adjusted to support the string-based values from parsing the DML operations.  

In addition, the new config option `log.mining.dml.parser` allows toggling between parsers used, defaulting to the newer parser as the default.  Several new JMX metrics were added to capture the total time spent initializing a mining session by Oracle, the time spent calling `ResultSet#next` in the result-set processor along with the time spent executing the parser stage for each DML operation.  When the connector is stopped, the logs will output all metrics including the number of processed DML operations so we can deduce the number of events/sec metrics from those values.

The one thing I do think is important to point out right now is looking at the DEBUG output from `LogMinerQueryResultProcessor` after each batch of results are processed.  I added a calculation that tracks the number of rows read from the result-set.  This is important because its not uncommon to get a result-set with thousands (sometimes hundreds of thousands) of rows and we only technically process a smaller subset of those rows.  Most of these extra rows come from the fact we have to process every single commit or rollback, which may be related to internal transactions or operations generated by the database that aren't related to any transaction or user operation we're interested in.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -191,6 +191,15 @@
             .withDescription("Complete JDBC URL as an alternative to specifying hostname, port and database provided "
                     + "as a way to support alternative connection scenarios.");
 
+    public static final Field LOG_MINING_DML_PARSER = Field.create("log.mining.dml.parser")
```
gunnarmorling, line 194: Let's make it an _internal_ option. Ideally, it shouldn't be needed, and we can refer users to it only in case they have issues with the new one.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
```diff
@@ -151,6 +182,7 @@ public ValueConverter converter(Column column, Field fieldDefn) {
             case Types.NUMERIC:
                 return getNumericConverter(column, fieldDefn);
             case Types.FLOAT:
+                // todo: DBZ-3078 LogMiner used getFloatConverter here, review impact.
```
gunnarmorling, line 185: Are you planning to address these todos as part of this PR, or as a follow-up?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerDmlParser.java
```diff
@@ -0,0 +1,473 @@
...
+            switch (sql.charAt(0)) {
+                case 'i':
+                    return parseInsert(sql);
+                case 'u':
+                    return parseUpdate(sql);
+                case 'd':
+                    return parseDelete(sql);
+            }
+        }
+        return null;
```
gunnarmorling, line 52: Should we throw an exception rather? What did the JsqlParser impl do in this case?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerDmlParser.java
```diff
@@ -0,0 +1,473 @@
...
+import io.debezium.connector.oracle.logminer.valueholder.LogMinerColumnValue;
+import io.debezium.connector.oracle.logminer.valueholder.LogMinerColumnValueImpl;
+import io.debezium.connector.oracle.logminer.valueholder.LogMinerDmlEntry;
+import io.debezium.connector.oracle.logminer.valueholder.LogMinerDmlEntryImpl;
+import io.debezium.data.Envelope;
+
+/**
+ * @author Chris Cranford
+ */
+public class LogMinerDmlParser {
```
gunnarmorling, line 22: Can you add a description to the class comment, in particular giving examples for each of the three statement types.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerDmlParser.java
```diff
@@ -0,0 +1,473 @@
...
+        boolean inColumnName = true;
+        boolean inColumnValue = false;
+        boolean inQuote = false;
+        boolean inSingleQuote = false;
+
+        // verify entering where-clause
+        if (!sql.substring(start, start + 7).equals(WHERE)) {
+            throw new DebeziumException("Failed to parse DML: " + sql);
+        }
+        start += WHERE.length();
```
gunnarmorling, line 386: Let's extract constants for all known lengths like this.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerMetricsMXBean.java
```diff
@@ -147,6 +147,21 @@
      */
     long getNetworkConnectionProblemsCounter();
 
+    /**
+     * @return the total number of milliseconds used to parse DDL/DML statements
+     */
+    long getTotalParseTime();
```
gunnarmorling, line 153: Those are the exposed MBean attribute names, correct? If so, let's encode the UOM name in the name, as we do it elsewhere (e.g. "SnapshotDurationInSeconds"). The reason being we cannot expose a proper type like `Duration`,  and a consumer should have no ambiguity over the UOM.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerQueryResultProcessor.java
```diff
@@ -35,10 +38,11 @@
  */
 class LogMinerQueryResultProcessor {
 
-    private final ChangeEventSource.ChangeEventSourceContext context;
+    private final ChangeEventSourceContext context;
     private final LogMinerMetrics metrics;
     private final TransactionalBuffer transactionalBuffer;
-    private final SimpleDmlParser dmlParser;
+    private final SimpleDmlParser legacyDmlParser;
```
gunnarmorling, line 44: Couldn't there be an interface for the two implementations, and there'd only be a single field here, of this type?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
```diff
@@ -61,14 +61,14 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(LogMinerStreamingChangeEventSource.class);
 
+    // TODO: change this to use MAX_QUEUE_SIE as the default
```
gunnarmorling, line 64: Log an issue for tracking?

gunnarmorling (commented):
@Naros, a few comments inline. I think it's good overall, thanks a lot!

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerDmlParser.java
Naros: So the other parser implementation was designed to catch any parse exception, log an error, and then return `null` which caused the processing loop to skip the event and move onto the next event as if nothing happened.  Would you agree we probably should throw some type of `DmlParserException` which would cause the connector task to fail?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
Naros: When you say _internal_, what specifically should I change here outside of not documenting this option?

Naros: Ah; nvm, you likely mean `Field.createInternal`.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
Naros: I am, they'll be fixed in the next commit I'm working on.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
Naros: I'm actually going to make this change as a part of this PR.  In fact, it should be sufficient enough to remove the constant entirely and replace it with `connectorConfig.getMaxQueueSize()`.

Naros:
Quick status update:

* `StreamingDatatypesIT` test is no longer skipped for LogMiner & now passes
  * Identified several problems with the converters which have been resolved.
  * Consolidated `OracleChangeRecordValueConverter` into `OracleValueConverters`
* Metrics
  * Metric unit is now explicit as part of the method name.
  * Added several new metrics, `TotalProcessedRows` and `TotalResultSetNextTimeInMilliseconds`
  * Corrected calculation for some metrics, e.g. total process time excludes pauses.
* Configuration
  * Parser configuration is now internal as requested.
* Parser
  * Legacy and new parser now implement a common interface
  * DmlParserException is thrown when both parsers fail to parse DML rather than returning `null`.  
  * Transactions no longer skipped upon parse errors.
  * New Parser now supports `Unsupported Type` column values, often returned for spatial data types.
* End-To-End JUnit benchmark

Right now there are 2 anomalies in the end-to-end benchmark and once those are resolved, I'll push the commit.  I just wanted to give a heads up of what to expect soon.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
gunnarmorling: Yes, exactly. Are you planning to make this change?

gunnarmorling (commented):
@Naros, added a few more comments. Seems like an earlier push is missing still?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
```diff
@@ -107,7 +137,7 @@ private SchemaBuilder getNumericSchema(Column column) {
                 else if (width < 5) {
                     return SchemaBuilder.int16();
                 }
-                // todo: DBZ-137 this was changed and caused issues with datatype tests, reverted for now.
+                // todo: DBZ-3078 This was reverted as of DBZ-137 but LogMiner streaming used this, review impact.
```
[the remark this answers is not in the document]
Naros: It's already resolved in the changes I am testing locally,

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
```diff
@@ -121,6 +151,7 @@ else if (width < 19) {
             return super.schemaBuilder(column);
         }
         else {
+            // todo: DBZ-3078 LogMiner streaming assumed VariableScaleDecimal.builder(), review impact.
```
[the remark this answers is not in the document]
Naros: Already resolved, see above comment.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerDmlParser.java
```diff
@@ -0,0 +1,473 @@
...
+/**
+ * @author Chris Cranford
+ */
+public class LogMinerDmlParser {
+
+    private static final String SINGLE_QUOTE = "'";
+    private static final String NULL = "NULL";
+    private static final String INSERT_INTO = "insert into ";
+    private static final String UPDATE = "update ";
+    private static final String DELETE = "delete ";
```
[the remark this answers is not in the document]
Naros: Fixed as well in the local changes I'm testing.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerDmlParser.java
```diff
@@ -0,0 +1,473 @@
...
+
+        // parse table
+        index = parseTableName(sql, index);
+
+        // capture column names
+        List<String> columnNames = new ArrayList<>();
+        index = parseColumnListClause(sql, index, columnNames);
+
+        // capture values
+        List<String> columnValues = new ArrayList<>();
```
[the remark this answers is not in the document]
Naros: Good point, I'll make this adjustment.

gunnarmorling:
Hey @AndreyIg, howdy? Thought this PR might be of interest for you. It's a new DML parser implementation which is substantially faster than the previous one. @Naros can share exact numbers, but he's reporting ~900K/sec parses for INSERTs with 10 columns for instance. So pretty sure the parser won't be the bottleneck any longer once this is in. We're planning to roll this out in next week's 1.5.0.Beta1.

debezium-microbenchmark-oracle/src/main/java/io/debezium/performance/connector/oracle/parser/LogMinerDmlParserPerf.java
```diff
@@ -0,0 +1,116 @@
...
+
+        private String insertStatement() {
+            final StringBuilder sb = new StringBuilder("insert into \"DEBEZIUM\".\"TEST\"(\"ID\"");
+            for (int i = 0; i < columnCount; ++i) {
+                sb.append(",").append("\"C").append(i).append("\"");
+            }
+            sb.append(") values (");
+            sb.append("'").append(1).append("'");
+            for (int i = 0; i < columnCount; ++i) {
+                sb.append(",").append("'V").append(i).append("'");
```
gunnarmorling, line 61: Values probably should be a bit longer to be closer to what you'd actually have. Like some strings with 50 char or so.

AndreyIg:
thank you for the very interesting info, guys.

gunnarmorling:
@Naros, could you rebase this one? And anything else you wanted to add? Or should we go for it?

Naros:
@gunnarmorling Rebased & this is ready to go.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/parser/DmlParser.java
```diff
@@ -0,0 +1,27 @@
...
+
+import io.debezium.connector.oracle.logminer.valueholder.LogMinerDmlEntry;
+import io.debezium.relational.Tables;
+
+/**
+ * Contract for a DML parser for LogMiner.
+ *
+ * @author Chris Cranford
+ */
+public interface DmlParser {
```
gunnarmorling, line 16: Great!

gunnarmorling (approved):
Woohoo!

gunnarmorling:
Aaand merged. Great piece of work, @Naros! Big thank you and congrats for driving this home!

PR #254, opened 2021-02-12
DBZ-3090 Fix emitting transaction end event for LogMiner

Naros:
https://issues.redhat.com/browse/DBZ-3090

The outstanding question mentioned in the associated issue is whether or not we believe we need to rework the commit handler logic used by the commit-thread executor.  For now, this PR simply addresses making sure that the transaction-end event is emitted when the transaction is committed.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/TransactionalBuffer.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/TransactionalBuffer.java
@@ -15,45 +15,35 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.ArrayBlockingQueue;
-import java.util.concurrent.BlockingQueue;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.ThreadPoolExecutor;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
-import java.util.function.Supplier;
+import java.util.concurrent.atomic.AtomicLong;
 
 import org.apache.kafka.connect.errors.DataException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import io.debezium.annotation.NotThreadSafe;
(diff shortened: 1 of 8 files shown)


Naros:
I decided in the second commit to focus on connector consistency and removed the separate commit thread entirely and simply do the commit in the main connector thread.  We may find we need to be multi-threaded in the future but for now I'd rather keep it simple, profile, and then focus on re-architecting where needed.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/TransactionalBuffer.java
```diff
@@ -187,39 +178,43 @@ boolean commit(String transactionId, Scn scn, OracleOffsetContext offsetContext,
...
-                metrics.setCommitQueueCapacity(commitQueueCapacity.get());
-                metrics.setLastCommitDuration(Duration.between(start, Instant.now()).toMillis());
-                taskCounter.decrementAndGet();
+        return true;
+    }
+
+    private void commit(ChangeEventSource.ChangeEventSourceContext context, OracleOffsetContext offsetContext, Instant start,
+                        List<CommitCallback> commitCallbacks, Timestamp timestamp, Scn smallestScn, Scn scn, EventDispatcher<?> dispatcher) {
+        try {
+            int counter = commitCallbacks.size();
```
gunnarmorling, line 189: Not part of this diff -- but take a look at this method, we do first a get() and then a remove() call on the `transactions` map. I'd expect the latter suffices?

gunnarmorling (approved):
LGTM, one small suggestion inline.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/TransactionalBuffer.java
Naros: Nice catch, method cleaned up and I also took the opportunity to remove `largestScn` from the buffer as was superfluous and no longer necessary.

gunnarmorling (approved):
LGTM, merging. Thx, @Naros!

PR #255, opened 2021-02-15
DBZ-3001 Control MAX Scn for LogMiner based on Oracle version

Naros:
https://issues.redhat.com/browse/DBZ-3001

This is an initial pass at providing a MAX SCN calculation based on the Oracle database version.  We've seeded the MAX values returned based on data gathered either through my testing of Oracle 12.2 and 19.3 as well user input for Oracle 19.6.

@ant0nk @mpermar if either of you can pull down this PR into your local branch and test it, I would be most grateful.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
@@ -56,10 +56,13 @@ public class OracleConnection extends JdbcConnection {
     /**
      * The database version.
      */
-    private OracleDatabaseVersion databaseVersion;
+    private final OracleDatabaseVersion databaseVersion;
 
     public OracleConnection(Configuration config, Supplier<ClassLoader> classLoaderSupplier) {
         super(config, resolveConnectionFactory(config), classLoaderSupplier);
+
+        this.databaseVersion = resolveOracleDatabaseVersion();
+        LOGGER.info("Database Version: {}", databaseVersion.getBanner());
     }
 
     public void setSessionToPdb(String pdbName) {
@@ -107,10 +110,6 @@ public class OracleConnection extends JdbcConnection {
     }
 
     public OracleDatabaseVersion getOracleVersion() {
-        if (databaseVersion == null) {
-            databaseVersion = resolveOracleDatabaseVersion();
-            LOGGER.info("Database Version: {}", databaseVersion.getBanner());
-        }
         return databaseVersion;
     }
 
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
@@ -53,6 +53,11 @@ public class OracleConnection extends JdbcConnection {
      */
     private static final Field URL = Field.create("url", "Raw JDBC url");
 
+    /**
+     * The database version.
+     */
+    private OracleDatabaseVersion databaseVersion;
+
     public OracleConnection(Configuration config, Supplier<ClassLoader> classLoaderSupplier) {
         super(config, resolveConnectionFactory(config), classLoaderSupplier);
     }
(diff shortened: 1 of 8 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
```diff
@@ -101,6 +106,60 @@ public void resetSessionToCdb() {
         }
     }
 
+    public OracleDatabaseVersion getOracleVersion() {
```
gunnarmorling, line 109: Is there a specific reason for doing this lazily? If not, let's do this in the constructor and make `databaseVersion` final.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
```diff
@@ -101,6 +106,60 @@ public void resetSessionToCdb() {
...
+                    return null;
+                });
+            }
+        }
+        catch (SQLException e) {
+            throw new RuntimeException("Failed to resolve Oracle database version", e);
+        }
+
+        if (versionStr == null) {
+            throw new RuntimeException("Failed to resolve Oracle database version");
```
gunnarmorling, line 157: So there's no way we can keep running without the version?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
```diff
@@ -124,87 +123,84 @@ public void execute(ChangeEventSourceContext context) {
...
-                startScn = offsetContext.getScn();
-                createFlushTable(connection);
-
-                if (!isContinuousMining && startScn < getFirstOnlineLogScn(connection, archiveLogRetention)) {
-                    throw new DebeziumException(
-                            "Online REDO LOG files or archive log files do not contain the offset scn " + startScn + ".  Please perform a new snapshot.");
-                }
+            if (!isContinuousMining && startScn < getFirstOnlineLogScn(jdbcConnection, archiveLogRetention)) {
+                throw new DebeziumException(
+                        "Online REDO LOG files or archive log files do not contain the offset scn " + startScn + ".  Please perform a new snapshot.");
```
gunnarmorling, line 135:
```suggestion
                        "Online redo log files or archive log files do not contain the offset scn " + startScn + ".  Please perform a new snapshot.");
```

gunnarmorling (commented):
@Naros, I guess it's fine, but I'm a bit disconnected from the details tbh. Why is it we do need to know that max SCN in the first place? And yes, would be nice to get some feedback from testing by @ant0nk and @mpermar.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
Naros: No specific reason, I'll move it.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
Naros: LogMinerHelper methods that deal with redo and archive logs depend on very specific magic SCN values that are derived based on the current database version.  If we return a default version here, this would cause those methods that depend on this specific magic number to lead to the notorious `missing logfile` exception we've been trying to avoid.

I can look into moving away from these magic SCN value checks entirely, but we'd need to rework the code around fetching and working with both archive and redo logs, i.e. knowing which ones to add to the session, etc.

gunnarmorling: Ok, I see. I can't really tell whether that's a feasible thing to look into or not. If you think it is, can you log it in Jira as a follow-up please, and we go with what we have for now.

gunnarmorling:
@Naros, seems it's ready to go from your PoV? Still curious re this:

> Why is it we do need to know that max SCN in the first place?

Naros:
> Why is it we do need to know that max SCN in the first place?

At least according to the code and the Oracle table definition the {{NEXT_CHANGE#}} column can be `NULL` and in this case the code replaces this `NULL` with the max SCN so that later code does not always need to perform null-value checks.  I was going to spend a small window after today's meeting to see if I could quickly remove this substitution practice but it requires a good understanding of several methods first.

Naros:
Just a quick status update, I think I've been able to integrate the `Scn` domain object into this "max SCN" usage.  I need to rerun the tests & verify there are no regressions, but if this works then technically we could return the lowest known supported version when we cannot detect the database version as a fallback solution and that would "just work".

Naros:
So there are some regressions I need to investigate, so I added a follow-up [DBZ-3145](https://issues.redhat.com/browse/DBZ-3145).  In the meantime, we can proceed w/what we have.

gunnarmorling:
So to be sure, you're saying this one can be merged, correct?

Naros:
Indeed

gunnarmorling:
Done, thx a lot!

PR #256, opened 2021-02-19
DBZ-3009 Support multiple schemas with Oracle LogMiner

Naros:
https://issues.redhat.com/browse/DBZ-3009

This PR removes the `database.schema` configuration option and now supports multiple schemas based on the include/exclude lists provided by the common connector configurations.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
@@ -76,14 +76,6 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
             .withDescription("Name of the pluggable database when working with a multi-tenant set-up. "
                     + "The CDB name must be given via " + DATABASE_NAME.name() + " in this case.");
 
-    public static final Field SCHEMA_NAME = Field.create(DATABASE_CONFIG_PREFIX + "schema")
-            .withDisplayName("Schema name")
-            .withType(Type.STRING)
-            .withWidth(Width.MEDIUM)
-            .withImportance(Importance.HIGH)
-            .withValidation(OracleConnectorConfig::validateDatabaseSchema)
-            .withDescription("Name of the connection user to the database ");
-
     public static final Field XSTREAM_SERVER_NAME = Field.create(DATABASE_CONFIG_PREFIX + "out.server.name")
             .withDisplayName("XStream out server name")
             .withType(Type.STRING)
@@ -314,7 +306,6 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
             Heartbeat.HEARTBEAT_TOPICS_PREFIX,
             TABLENAME_CASE_INSENSITIVE,
             ORACLE_VERSION,
-            SCHEMA_NAME,
             CONNECTOR_ADAPTER,
             LOG_MINING_STRATEGY,
             SNAPSHOT_ENHANCEMENT_TOKEN,
@@ -341,7 +332,6 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
 
     private final boolean tablenameCaseInsensitive;
     private final OracleVersion oracleVersion;
-    private final String schemaName;
     private final Tables.ColumnNameFilter columnFilter;
     private final HistoryRecorder logMiningHistoryRecorder;
     private final Configuration jdbcConfig;
@@ -374,7 +364,6 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
         this.snapshotMode = SnapshotMode.parse(config.getString(SNAPSHOT_MODE));
         this.tablenameCaseInsensitive = config.getBoolean(TABLENAME_CASE_INSENSITIVE);
         this.oracleVersion = OracleVersion.parse(config.getString(ORACLE_VERSION));
-        this.schemaName = toUpperCase(config.getString(SCHEMA_NAME));
         String blacklistedColumns = toUpperCase(config.getString(RelationalDatabaseConnectorConfig.COLUMN_BLACKLIST));
         this.columnFilter = getColumnNameFilter(blacklistedColumns);
         this.logMiningHistoryRecorder = resolveLogMiningHistoryRecorder(config);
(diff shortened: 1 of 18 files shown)


Naros:
The new tests added for `OracleConnectorFilterIT` fail for Xstream, investigating.

Naros:
Tests fixed, should be ready for review.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerQueryResultProcessor.java
```diff
@@ -156,6 +157,11 @@ int processResult(ResultSet resultSet) {
 
             // DML
             if (operationCode == RowMapper.INSERT || operationCode == RowMapper.DELETE || operationCode == RowMapper.UPDATE) {
+                final TableId tableId = RowMapper.getTableId(connectorConfig.getCatalogName(), resultSet);
```
gunnarmorling, line 160: So we are getting the entries for _all_ tables and only filter out later on, correct? Did you consider crafting a WHERE class based on the filter configuration, so to select only the captured schemas (I think then we'd effectively be doing the same as before)?

gunnarmorling: Or maybe not, given the changes further below? Why do we do this filtering after the fact here?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
```diff
@@ -335,21 +412,4 @@ public static long parseRetentionFromName(String historyTableName) {
                 Integer.parseInt(tokens[6])); // minutes
         return Duration.between(recorded, LocalDateTime.now()).toHours();
     }
-
-    /**
-     * This method builds table_name IN predicate, filtering out non whitelisted tables from Log Mining.
-     * It limits joining condition over 1000 tables, Oracle will throw exception in such predicate.
-     * @param tables white listed table names
-     * @return IN predicate or empty string if number of whitelisted tables exceeds 1000
```
gunnarmorling, line 343: Do we still have this consideration of the max IN size somewhere?

gunnarmorling (commented):
I think it's good, one question inline. Thx!

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
Naros: No.  We use the Oracle `REGEXP_LIKE` operator that takes each of the include/exclude list regex patterns and applies a single predicate to the where-clause for each pattern.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerQueryResultProcessor.java
Naros: This is to handle a situation where the configuration does not apply any include/exclude lists but we need to still filter out the events we process for objects only in the configured `database.pdb.name` catalog.  I don't believe there was a way to do this directly with the SQL query, but I can take one last pass to see if we can.

Naros: So after reviewing the code, the missing link was `SystemTablesPredicate`.  This filter is used in order for the programmatic code to exclude certain tables that are in known excluded schemas such as `SYS`.  The dynamic nature of the LogMiner query was not taking into account these "excluded" schemas.

For now, I'll move these tables defined in `SystemTablesPredicate` likely to a `SystemTables` class that both the predicate and the `SqlUtils` class used by LogMiner can be based on.  This would allow us to maintain this list once and should solve the concern with the additional filter call you referenced above.  Once I've made the changes & verified the tests pass, I'll push the rebase & this last change.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
gunnarmorling: Wow, I see. I'm a tad concerned there may be subtle differences in the regexp engine semantics between Java and Oracle's, potentially resulting in inconsistencies between connectors. So is filter handling for Oracle then _exclusively_ done via that database function?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -131,7 +125,7 @@
 
     public static final Field LOG_MINING_STRATEGY = Field.create("log.mining.strategy")
             .withDisplayName("Log Mining Strategy")
-            .withEnum(LogMiningStrategy.class, LogMiningStrategy.CATALOG_IN_REDO)
+            .withEnum(LogMiningStrategy.class, LogMiningStrategy.ONLINE_CATALOG)
```
gunnarmorling, line 128: Didn't we want to use the save but slow option by default?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
```diff
@@ -199,34 +199,124 @@ static String startLogMinerStatement(Long startScn, Long endScn, OracleConnector
...
+    static String logMinerContentsQuery(OracleConnectorConfig connectorConfig, String logMinerUser) {
+        StringBuilder query = new StringBuilder();
+        query.append("SELECT SCN, SQL_REDO, OPERATION_CODE, TIMESTAMP, XID, CSF, TABLE_NAME, SEG_OWNER, OPERATION, USERNAME ");
+        query.append("FROM ").append(LOGMNR_CONTENTS_VIEW).append(" ");
+        query.append("WHERE OPERATION_CODE IN (1,2,3,5) ");
+        query.append("AND SCN >= ? ");
+        query.append("AND SCN < ? ");
+        query.append("AND TABLE_NAME != '").append(LOGMNR_FLUSH_TABLE).append("' ");
+
+        List<String> excludedSchemas = OracleConnectorConfig.getExcludedSchemaNames();
```
gunnarmorling, line 231: Worth some commenting here what's going on.

gunnarmorling (commented):
Two more comments inline. Is it good to go after that from your PoV, @Naros?

gunnarmorling:
Ok, so as discussed on the call: let's documented that only POSIX regular expressions should be used with LogMiner.

gunnarmorling:
Also schema filter options aren't documented yet at all.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
Naros: Yes we did, I'll fix.

Naros:
Documented the changes in https://github.com/debezium/debezium/pull/2155

Naros:
@gunnarmorling @jpechane I believe this should now be ready to go with the documentation changes & the suggested changes from the last review.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
Naros: For LogMiner during streaming, yes filtering is handled exclusively by the database query itself.  For all other portions of the code, including Xstreams or during Snapshotting, this is handled in Java exclusively.

gunnarmorling:
Applied, thx!

PR #258, opened 2021-02-19
DBZ-3110 Introduce RequireDatabaseOption test annotation

Naros:
https://issues.redhat.com/browse/DBZ-3110


gunnarmorling (approved):
Worth moving to core repo eventually? But LGTM in any case, we can do that later on.

PR #259, opened 2021-02-19
DBZ-3153 Capture LogMiner parameters when LogMiner session fails to start

Naros:
https://issues.redhat.com/browse/DBZ-3153


gunnarmorling:
Applied, thx!

PR #260, opened 2021-02-19
DBZ-3151 Properly handle system-generated hidden column unique indexes

Naros:
https://issues.redhat.com/browse/DBZ-3151

This PR depends on https://github.com/debezium/debezium/pull/2150.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
@@ -20,6 +20,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.Supplier;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


Naros:
Per a recent conversation, there may be another use case with indices we need to consider, columns named `SYS_NCnnnnn$`.  I've requested the table DDL in the attached issue and once we get that, I'll add a test & verify the necessary fix.

foryou7242:
hello i was testing function based index 


 CREATE UNIQUE INDEX "ADM"."COUPON_PK" ON "ADM"."COUPON" ("NUM", "COUPON_AREA_TP", "COUPON_NO" DESC, "COUPON_PBLCT_NO", "SEQ")

Naros:
Updated test to use a pattern to detect `SYS_NC_OID$`, `SYS_NC_ROWINFO$`, and `SYS_NCnnnnn$` naming conventions.  These should all be ignored as they're considered auto-generated by Oracle and aren't a logical column in the table.

gunnarmorling (approved):

PR #265, opened 2021-02-23
1.5.b1 test fixes

Naros:
@gunnarmorling This is the PR I discussed on our call today.  This addresses the JAXB test case failure as well as a small NPE that I uncovered with the recent DBZ-3151 changes.


gunnarmorling:
Thx!

PR #266, opened 2021-02-23
DBZ-3167 Removed column.blacklist, cleaned up config definition

Naros:
https://issues.redhat.com/browse/DBZ-3167

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
@@ -17,6 +17,7 @@ import org.apache.kafka.common.config.ConfigDef.Type;
 import org.apache.kafka.common.config.ConfigDef.Width;
 
 import io.debezium.config.CommonConnectorConfig;
+import io.debezium.config.ConfigDefinition;
 import io.debezium.config.Configuration;
 import io.debezium.config.EnumeratedValue;
 import io.debezium.config.Field;
@@ -29,7 +30,6 @@ import io.debezium.connector.oracle.xstream.LcrPosition;
 import io.debezium.connector.oracle.xstream.OracleVersion;
 import io.debezium.document.Document;
 import io.debezium.function.Predicates;
-import io.debezium.heartbeat.Heartbeat;
 import io.debezium.jdbc.JdbcConfiguration;
 import io.debezium.relational.ColumnId;
 import io.debezium.relational.HistorizedRelationalDatabaseConnectorConfig;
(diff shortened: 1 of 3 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -472,10 +441,6 @@ public OracleVersion getOracleVersion() {
         return oracleVersion;
     }
 
-    public Tables.ColumnNameFilter getColumnFilter() {
```
gunnarmorling, line 475: How does the column filter work then now? Is it inherited from the framework? Is there a test for the Oracle connector for this?

gunnarmorling (commented):
One question inline.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
Naros: Yes the filter is part of the common framework; however I do need to add a test for this.  I thought this was tested in the core repository but looking at the connectors there they each have at least one test for this so I'll add one here for consistency.

gunnarmorling:
Yes, we should have connector-specific tests for such shared functionality.
Ideally, we'd have a notion of a TCK which can be run against all the
(relational) connectors, but we're not there yet.

>

Naros:
@gunnarmorling Tests added.

Naros:
I need to add `column.include/exclude.list` to the docs as those aren't present.

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/OracleConnectorFilterIT.java
```diff
@@ -453,6 +477,140 @@ private void shouldApplySchemaAndTableExclusionsConfiguration(boolean useLegacyO
         }
     }
 
+    private void shouldApplyColumnInclusionsConfiguration(boolean useLegacyOption) throws Exception {
+        Field option = OracleConnectorConfig.COLUMN_INCLUDE_LIST;
+        if (useLegacyOption) {
```
gunnarmorling, line 482: So if we never documented these legacy option names as per what your say, I'd suggest let's simply remove them altogether. Chances anyone used them are really low then, and given the connector is incubating, the remaining risk seems acceptable to me.

Naros: Fixed, tests now only use the include/exclude option names for column filtering.

gunnarmorling (approved):

gunnarmorling:
Awesome; pretty cool to see how we become such features essentially "for free" by means of the framework 🎉 .

PR #269, opened 2021-02-23
Fix Benchmark for 1.5.0.Beta1

Naros:
@jpechane This is the small fixes to get the benchmark module to compile.
Sorry for missing these.


jpechane:
@Naros Applied, thanks for the fast fix!

PR #277, opened 2021-03-02
DBZ-3208 Don't emit NUMBER(1) as boolean types

Naros:
https://issues.redhat.com/browse/DBZ-3208

* NUMBER(1) column types emit as numeric
* Added OOTB converter, NumberOneToBooleanConverter


gunnarmorling:
Nice one, thx!

PR #278, opened 2021-03-02
DBZ-3154 Update test suite to run in CDB & non-CDB modes.

Naros:
https://issues.redhat.com/browse/DBZ-3154

This should be merged in conjunction with https://github.com/debezium/oracle-vagrant-box/pull/18 as the changes in the helper class reflects the various script changes in the vagrant box repository.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
@@ -358,7 +358,7 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
     private final LogMiningDmlParser dmlParser;
 
     public OracleConnectorConfig(Configuration config) {
-        super(OracleConnector.class, config, config.getString(SERVER_NAME), new SystemTablesPredicate(), x -> x.schema() + "." + x.table(), true);
+        super(OracleConnector.class, config, config.getString(SERVER_NAME), new SystemTablesPredicate(config), x -> x.schema() + "." + x.table(), true);
 
         this.databaseName = toUpperCase(config.getString(DATABASE_NAME));
         this.pdbName = toUpperCase(config.getString(PDB_NAME));
@@ -709,9 +709,24 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
      */
     private static class SystemTablesPredicate implements TableFilter {
 
+        private final Configuration config;
+
+        SystemTablesPredicate(Configuration config) {
+            this.config = config;
+        }
+
         @Override
         public boolean isIncluded(TableId t) {
-            return !OracleConnectorConfig.getExcludedSchemaNames().contains(t.schema().toLowerCase());
+            return !isExcludedSchema(t) && !isFlushTable(t);
+        }
+
+        private boolean isExcludedSchema(TableId id) {
+            return OracleConnectorConfig.getExcludedSchemaNames().contains(id.schema().toLowerCase());
+        }
+
+        private boolean isFlushTable(TableId id) {
+            final String schema = config.getString(USER);
+            return id.table().equalsIgnoreCase("LOG_MINING_FLUSH") && id.schema().equalsIgnoreCase(schema);
         }
     }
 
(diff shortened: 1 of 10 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -709,9 +709,24 @@ public static LogMiningDmlParser parse(String value, String defaultValue) {
...
+        }
+
         @Override
         public boolean isIncluded(TableId t) {
-            return !OracleConnectorConfig.getExcludedSchemaNames().contains(t.schema().toLowerCase());
+            return !isExcludedSchema(t) && !isFlushTable(t);
+        }
+
+        private boolean isExcludedSchema(TableId id) {
+            return OracleConnectorConfig.getExcludedSchemaNames().contains(id.schema().toLowerCase());
```
gunnarmorling, line 724: Let's make the list backing `getExcludedSchemaNames()` a constant instead of the local `Arrays.asList()` call.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -709,9 +709,24 @@ public static LogMiningDmlParser parse(String value, String defaultValue) {
...
+        }
+
+        private boolean isFlushTable(TableId id) {
+            final String schema = config.getString(USER);
+            return id.table().equalsIgnoreCase("LOG_MINING_FLUSH") && id.schema().equalsIgnoreCase(schema);
```
gunnarmorling, line 729: Let's refer to the constant in `SqlUtils`.

gunnarmorling (commented):
@Naros, two comments inline. Also, how does the test suite gets run with CDB vs. PDB modes? Is some update to the README due? Or even update to CI config to test both modes? // CC @jpechane

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/logminer/SqlUtilsTest.java
```diff
@@ -321,11 +321,10 @@ public void shouldDetectConnectionProblems() {
     private String resolveLogMineryContentQueryFromTemplate(String schemaReplacement, String tableReplacement) {
         String query = LOG_MINER_CONTENT_QUERY_TEMPLATE;
 
-        List<String> excludedSchemas = OracleConnectorConfig.getExcludedSchemaNames();
-        if (!excludedSchemas.isEmpty()) {
+        if (OracleConnectorConfig.EXCLUDED_SCHEMAS.length > 0) {
             StringBuilder systemPredicate = new StringBuilder();
             systemPredicate.append("AND SEG_OWNER NOT IN (");
-            for (Iterator<String> i = excludedSchemas.iterator(); i.hasNext();) {
+            for (Iterator<String> i = Arrays.stream(OracleConnectorConfig.EXCLUDED_SCHEMAS).iterator(); i.hasNext();) {
```
gunnarmorling, line 327: Tiny nit: couldn't `EXCLUDED_SCHEMAS` be an (unmodifiable) list, and you simply use a for each loop? Or is there any specific need for the iterator here?

gunnarmorling (commented):
@Naros one follow-up :) And any comments on my other questions (README, CI)?

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/logminer/SqlUtilsTest.java
Naros: No particular reason, I can certainly change that.

Naros:
>  Also, how does the test suite gets run with CDB vs. PDB modes? Is some update to the README due? Or even update to CI config to test both modes?

So the way to test the connector in CDB mode is just to run the test suite as you normally would.  If you provide the test suite with a `database.pdb.name` parameter with a non-empty value or if you don't specify this setting, the test suite will assume that it is in CDB mode just like it always had.  The caveat with the test suite is that if you specify `database.pdb.name` but with an empty value, the test suite has a special handling of this use case to trigger non-CDB mode.

That said, I'll add a note in the `README` that describes this as well.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -725,7 +727,7 @@ public boolean isIncluded(TableId t) {
         }
 
         private boolean isExcludedSchema(TableId id) {
-            return Arrays.stream(EXCLUDED_SCHEMAS).anyMatch(id.schema().toLowerCase()::equals);
+            return EXCLUDED_SCHEMAS.contains(id.schema().toLowerCase());
```
gunnarmorling, line 730: Thank you :)

gunnarmorling (approved):

